PR #2, opened 2018-07-13
Add a manifest for the ANTS Framework

knightsc:
The ANTS Framework helps set up ansible in pull mode to manage
a mac computer. This manifest has all keys available to configure
the ANTS Framework. Any value not set via profile will default to
the local ants.cfg file.

image

The ANTS Profile is a MCX profile that has the format similar to this

```
<key>mcx_preference_settings</key>
<dict>
    <key>main</key>
    <dict>
        <key>wait_interval</key>
        <integer>900</integer>
        <key>inventory_script</key>
        <string>inventory_default</string>
        <key>destination</key>
        <string>~root/.ants_playbook</string>
        <key>ssh_key</key>
        <string>/etc/ants/id_ants</string>
        <key>ssh_stricthostkeychecking</key>
        <false/>
        <key>git_repository</key>
        <string>https://github.com/ANTS-Framework/playbook.git</string>
        <key>log_dir</key>
        <string>/var/log/ants</string>
        <key>branch</key>
        <string>master</string>
        <key>ansible_pull_exe</key>
        <string></string>
        <key>ansible_playbook</key>
        <string>main.yml</string>
    </dict>
    <key>ad</key>
    <dict>
        <key>ldap_user</key>
        <string>ldap\changeme</string>
        <key>ldap_pw</key>
        <string>changeme</string>
        <key>ldap_host</key>
        <string>LDAP.PRETENDCORP.COM</string>
        <key>ldap_ou_computers</key>
        <string>DC=ldap,DC=pretendcorp,DC=com</string>
        <key>ldap_ou_groups</key>
        <string>OU=ants,OU=Groups,DC=ldap,DC=pretendcorp,DC=com</string>
        <key>cache_file</key>
        <string>/etc/ants/inventory_cache.json</string>
        <key>group_prefix</key>
        <string>ants-</string>
        <key>common_group</key>
        <string>ants-common</string>
        <key>common_ad_connected</key>
        <string>common-ad-connected</string>
    </dict>
</dict>
```

The consensus in the #ansible channel on the Mac Admins slack was that the github repo and strict host key checking flag should be required. I could not figure out how to move those two entries outside of the segmented control while still having them nested under either the `main` or `ad` keys. Additionally I wasn't sure how the image data worked but there is an image here that can be used inside ProfileCreator

https://avatars0.githubusercontent.com/u/31036978?s=200&v=4

### DIFF
--- a/ApplicationsManagedPreferences/ANTSConfigProfile.plist
+++ b/ApplicationsManagedPreferences/ANTSConfigProfile.plist
@@ -1,0 +1,425 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>pfm_app_documentation_url</key>
+	<string>https://ants-framework.github.io</string>
+	<key>pfm_app_url</key>
+	<string>https://github.com/ANTS-Framework/ants</string>
+	<key>pfm_description</key>
+	<string>ANTS Framework settings</string>
+	<key>pfm_domain</key>
+	<string>com.github.ants-framework</string>
+	<key>pfm_format_version</key>
+	<integer>1</integer>
+	<key>pfm_icon</key>
+	<data></data>
+	<key>pfm_last_modified</key>
+	<date>2018-07-12T11:54:43Z</date>
+	<key>pfm_subkeys</key>
+	<array>
+		<dict>
+			<key>pfm_default</key>
+			<string>Configures the ANTS framework configuration preferences</string>
+			<key>pfm_description</key>
+			<string>Description of the payload</string>
+			<key>pfm_name</key>
+			<string>PayloadDescription</string>
+			<key>pfm_title</key>
+			<string>Payload Description</string>
+			<key>pfm_type</key>
+			<string>string</string>
+		</dict>
+		<dict>
+			<key>pfm_default</key>
+			<string>ANTS Framework settings</string>
+			<key>pfm_description</key>
+			<string>Name of the payload</string>
+			<key>pfm_name</key>
+			<string>PayloadDisplayName</string>
+			<key>pfm_require</key>
+			<string>always</string>
+			<key>pfm_title</key>
+			<string>Payload Display Name</string>
+			<key>pfm_type</key>
+			<string>string</string>
+		</dict>
+		<dict>
+			<key>pfm_default</key>
+			<string>ants-framework</string>
+			<key>pfm_description</key>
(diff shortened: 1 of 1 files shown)


erikberglund:
I don't think there is a way currently to move the keys outside the segmented control when they are nested like this.

Previously when I tested this app in Objective-C I tried to make it exactly like ProfileManager where the interface could contain multiple different domains etc. And to do that I had a type of reference subkey that pointed to the actual domain and key and inserted it. Maybe a variation of that could be a solution to this problem, and then if they are referenced from inside the same manifest, don't show the original when it has a reference pointing to it.

Another solution is, of course, another key in the segmented control to say put the following keys in this array above or below itself.

Or another key on the subkey itself to say ignore the segmented control and put me above or below etc.

Do you have an idea for how it would make sense to write this type of thing in the manifest? What would be understandable and not unnecessarily create extra single-use keys?

A reference subkey might be useful in the future to support multiple domain manifests for example.

Regarding the icon, we could either just add an icon folder in the repository and the app pulls that together with the manifests. And/Or we could add the data key `pfm_icon` and just data encode the icon in the actual manifest. (I had this already implemented but removed it while cleaning some things). Both could be supported and the manifest could have priority.

Currently the framework handling the payloads is reading the payload domain and looking in it's resources folder (inside the folder for either application manifests or apple manifests etc) for an image with the same name as the domain.

knightsc:
When I was first working through understanding the manifest format I half expected anything that I marked as required to automatically show at the top of the form. This was mostly just due to browsing through the application and that's how they all displayed. I don't know that I would go that far and say anything marked required show automatically show at the top.

I do like the adding of having another key on the subkey itself so you could opt in to displaying it in a different manner. I think the advantage of that would be you could keep the main structure of all the subkeys matching the profile you want to create but just indicate it should be displayed in a somewhat different manner.

erikberglund:
Yes, it's true that required keys are put at the top when "disable optional keys" is enabled (it's the default). But when I implemented segmented controls, the keys had to stay in the same place to not make it weird to use, so all keys stay in place regardless of required or not. That is an easy change, to put the required keys on the top. That limits the layout options in another way but might be more obvious. I mean, if you want to create a layout like the main payload in profile manager where the outgoing and incoming mail server is required but under different tabs.

I could make a change to always put required keys on top regardless if there is a segmented control or not.

erikberglund:
It might become weird for conditional required keys to jump up and down, so a key to say explicitly might be better to keep consistency. We probably have to test what feels better.

erikberglund:
I've done a change that in segmented controls, if any key is marked in the manifest as require = always, then it will be put at the top. It's a quite specific change that matches this manifest, and I think when there are more I will probably revisit this and maybe add a setting or change the implementation so it doesn't interfere with the layout one wants. But, for this specific manifest, this works so I will put this in and we'll go from there. This is a baseline, like many other things where it has become impossible to know all the different configurations different domains want, Apple included.

PR #9, opened 2018-11-24
Fix minor spelling errors & Add 3 manifests

apizz:


apizz:
Also added AirServer and DetectX Swift application manifests

erikberglund:
Thank you. Sorry for the delay.

erikberglund:
@apizz I added icons for all new manifests. 

I also added the `pfm_hidden` key to the plist version key in the nomad shares manifest, as that key is required but also always 1 (at the moment) so it could just be removed form the GUI so a user can focus on the settings they should set.

It can be shown by showing hidden keys from the settings for the profile.

apizz:
Thanks for the icon updates and for the addition of the `pfm_hidden` key.  Will put that in my back pocket for next time.

PR #14, opened 2018-12-10
Update com.apple.dock.plist

qu1gl3s:
Addition of show-recents covering the 'Show recent applications in Dock' introduced in Mojave.


erikberglund:
Thank you.

PR #15, opened 2018-12-15
Add Platypus manifest & Add Plugins to Munki manifest

apizz:
In the munki manifest I added a new "Plugins" menu.  At the moment this only has S3-Auth (https://github.com/waderobson/s3-auth) preferences, but thought this structure would work for future plugin additions.

An important note, I'm not sure why, but the S3-Auth menu appears at the very bottom of the "General", "Managed Software Center", "URL", etc. menus with none of the keys.  However, "S3-Auth" and the proper keys are listed in Plugins.  I looked at my code and other examples but I can't seem to find why "S3-Auth" isn't only showing in Plugins.  Hope you might help on that front 😅

### DIFF
--- a/Manifests/ManagedPreferencesApplications/ManagedInstalls.plist
+++ b/Manifests/ManagedPreferencesApplications/ManagedInstalls.plist
@@ -16,7 +16,7 @@
 	<data>
 	</data>
 	<key>pfm_last_modified</key>
-	<date>2018-09-05T06:42:32Z</date>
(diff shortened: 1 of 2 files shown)


erikberglund:
No problem, it was just a typo you made with the key name, all was correct except for that!

erikberglund:
Forgot to add this to the update, but it's easily downloaded from within the app as usual.

apizz:
Much appreciated!

PR #16, opened 2018-12-19
Profile Manifest for Firefox Preferences

bp88:
Here is the first profile manifest for Firefox preferences based off the plist template https://github.com/mozilla/policy-templates/blob/master/mac/org.mozilla.firefox.plist. A couple of notes:
1. Verbiage was taken from a mix of the README on https://github.com/mozilla/policy-templates and the Windows GPO ADM definitions. I've replaced where applicable references to the word "policy" for "preference" since the target audience are macOS administrators.
2. I've made use of segmented controls to better categorize the long list of preferences. Those categories are:
General > General	
General > Add-ons	
General > Bookmarks	
General > Languages	
Network	
Privacy > General	
Privacy > Cookies	
Privacy > Permissions	
Security > General	
Security > Certificates	

3. The preference key `Proxy` has too variations in which other keys may require that other keys are selected. If you open up Network Settings in Firefox, you can see the variations and which fields are required depending on what you select. I was unable to find a way to make requirements here that made sense.

4. I cannot make sense of the preference key `SecurityDevices`. It is included but I'm not entirely sure whether it will work or not as the key:value names given do not make much sense in usage for a Configuration Profile.

5. There is an area for improvement in the preference key `Permissions` which has quite a few nested dictionaries. I was thinking of making a segmented control here, but open to feedback for better ease of use.

6. I've added pfm_title keys where it made sense for all the high level preference keys. There are situations where nested keys do not have pfm_title keys and could probably use them, but the descriptions should make it clear what the keys do.

7. Given that this is the first iteration I'm comfortable submitting, I've titled it Mozilla Firefox (Testing) so others can use it with knowledge that it hasn't been widely tested.

### DIFF
--- a/Manifests/ManagedPreferencesApplications/org.mozilla.firefox.plist
+++ b/Manifests/ManagedPreferencesApplications/org.mozilla.firefox.plist
@@ -1718,6 +1718,8 @@ For more information, see https://developer.mozilla.org/en-US/docs/Mozilla/Integ
 						<string>Optional. Select SOCKS Version.</string>
 						<key>pfm_name</key>
 						<string>SOCKSVersion</string>
+						<key>pfm_title</key>
+						<string>SOCKS Version</string>
 						<key>pfm_type</key>
 						<string>string</string>
 						<key>pfm_range_list</key>


erikberglund:
Firefox Changes:

Typo:

* Key `SOCKSVersion` was set to `Mode`, probably because of copy/past and forgot to change (what I often do).
	
Added Settings:

* Key `Mode` was required but had no `pfm_required` key, added that.
	
Descriptions:

* Removed all "Required." and "Optional." words from the descriptions for the UI. That is not needed as the app will handle if something is required. Those words are still left in the `pfm_description_reference` to a have a correct reference for anything no this app.
	
3)

* Moved keys `AutoLogin` and `UseProxyForDNS` as they are available to all except when proxy is set to none.
* Added `pfm_exclude` arrays to all keys that had conditions depending on the `Mode` key.
	
This results in that all keys can be enabled in the app, but only the relevant keys depending on the Mode selection will be exported.

There are other ways of handling this, first, if any of these keys are required if the Mode is set to something, that should be set as required on a condition instead of exclude on a condition.

There is also the option to hide/unhide keys from the GUI depending on the selection. I have not done that much to give users more visibility on what keys are available, but that is an option that could be used to make the GUI easier to configure and use. I don't think there is a generic manifest setting to do this at the moment but it could be added if needed.
	
So, the GUI is a little cluttered, but shows if the keys are enabled but wont be exported and why.
	
Ideally the yellow text that describes why it's excluded should maybe show what setting it expects instead of the current, but that would require a quite a large code work so haven't done that yet with the limited time I have.

4) 

I understand that you hade issues with this one, it requires both the Key and the Value of the dict to be user entered.
	
I have a configuration option for that, that I added now if you want to look at it.
	
And I have no input on if it's useful or valid, but with the config I added it can be entered correctly.

5)

Yes, when the nesting gets deeper it's hard to make a satisfying and clear UI. But, I have a key called `pfm_hidden` with an option of `container` that can hide "container" keys that themselves don't have any settings. In this case I added that key to the root dict, as it doesn't have to be shown in the GUI, and instead continued with your "X Permissions" structure, where I added a description to the Camera, Microphone dicts instead.
	
I think that makes it a little more clear. It's not perfect and a segmented control might be even better, but they will also become very nested.

6)

I renamed it Firefox to match the app name, even if you feel like there might be improvements it is easier to navigate and find when it's just the app name.
	
__

The key `RequestedLocales` looks great, but it says that you should be able to rearrange the languages in the order you want them, and the app can't do that right now. So, that is a feature needed to be implemented. But for now, this layout will work fine, but I think they will be in alphabetical order of enabled.

erikberglund:
Added the title and merged and updated the index. So now it should be visible to all clients.

PR #17, opened 2018-12-23
Preliminary Chrome Manifest Organization & other fixes

apizz:
There are a LOT of changes here and I've done a fair amount of testing as far as the organization and appearance of preferences goes.  I left a number of preferences in places that make sense to me, leaving others that don't listed under "Other", so I definitely need some testing & feedback here.

My decisions for organization followed these general rules:

1. Match the organizational categories for preferences as shown in the [Chrome documentation](https://dev.chromium.org/administrators/policy-list-3)
2. Organize preferences based on their organization in Google Chrome's settings (chrome://settings/)
3. Where preferences were left uncategorized, group them based on their relation to other preferences.  For example, I grouped the various Cookies, Images, JavaScript, etc. preferences in separate menu items under "Content Settings"
4. Dump everything else into an "Other" category (permanently, or with the hope that other kind souls have ideas for them)

Because of the current design where menus will not expand if the ProfileCreator application window width expands, or will not stack on top of another, I opted to build several menu lists under a given "primary" category.

Some other unrelated organizational changes:
- The `RegisteredProtocolHandlers` preference was missing several subkeys, which I added.
- Added specific `pfm_documentation_url` entries for each preference for easier reference back to Google's documentation.

Per https://github.com/erikberglund/ProfileCreator/issues/88, left `ManagedBookmarks` section commented out given current limitation in terms of array / dictionary depth.  As such, won't be able to add folders to `ManagedBookmarks`.

Menu Additions:
- General
    - General Menu List 1
        - Appearance
        - Search
        - On Startup
        - Printing
        - Updates
    - General Menu List 2
        - Autofill
        - Import
        - Google Cast
        - Cloud Print
        - Remote Access
- Extensions
    - Extensions Menu List 1
        - Reporting
        - Black/Whitelist
- Security
    - Security Menu List 1
        - Ads
        - History
        - Login / User
        - URLs
        - Plugins
        - Downloads
        - Networking
    - Security Menu List 2
        - Certs
        - Passwords
        - Proxy
        - HTTP Auth
        - Safe Browsing
        - Messaging
- Content Settings
    - Content Settings Menu List 1
        - Cookies
        - Images
        - JavaScript
        - Plugins
        - Popups
        - Web
    - Content Settings Menu List 2
        - Audio
        - Autoplay
        - Notifications
        - PDF
        - Video
- Other
    - Other Menu List 1
        - Disk Cache


bp88:
I like the categorization. It definitely shortens the long list if Chrome preferences one has to go through.

However, I think it can be a little confusing from a UX perspective that you have menus nested that do not change. For example, if you switch from:
"Content Settings > Cookies" to "Content Settings > Images", the second settings menu list "Content Settings > Audio" does not change.

I know you only did this because segmented controls do not currently resize. A better solution would be to have all the segmented controls laid out on the same level (e.g. Content Settings Menu List 1  and Content Settings Menu List 2 are on the same level) like you mentioned. However I believe that might require code changes to ProfileCreator as I don't think it currently can resize segmented controls to be on multiple lines.

I went ahead and opened a feature request for that to be implemented in ProfileCreator here: https://github.com/erikberglund/ProfileCreator/issues/100

apizz:
My Github n00bness is showing ... did not create separate branches for these different commits ... my apologies.  Will not make this same mistake moving forward!

apizz:
Closing this PR to create PR that only includes Chrome changes

PR #52, opened 2019-02-15

qu1gl3s:
Changes needed for the change of the Application ID for Microsoft AutoUpdate from the default MSau03 to MSau04 and the option of MSau03 for previous versions.


qu1gl3s:
Per https://github.com/erikberglund/ProfileManifests/issues/43

apizz:
Hey @qu1gl3s, can you please remove the updated index file from this PR?  Once we do the PR then we'll run the index update from the main repo.

I made this same mistake when working in git early on, but moving forward, please first create a branch off your master and then commit your changes there.  That way when you submit PRs you do it from that branch, rather than your fork master branch.

PR #58, opened 2019-03-01
Replaced incorrect payload identifier and type in com.apple.NSExtension.plist

mikedowler:


mikedowler:
Sorry - I'm new to git and had the wrong personal info present previously

Issue #43, opened 2019-02-14
Update Microsoft Office (com.microsoft.autoupdate2)


qu1gl3s:
Change Application ID from MSau03 to MSau04 as of version 4.7.1 of Microsoft AutoUpdate.

MSau03 for all versions prior to 4.7.1.

qu1gl3s:
Changes pending under https://github.com/erikberglund/ProfileManifests/pull/52

apizz:
Closed per #60

PR #60, opened 2019-03-05

qu1gl3s:
Changes needed for the change of the Application ID for Microsoft AutoUpdate from the default MSau03 to MSau04 and the option of MSau03 for previous versions.


qu1gl3s:
Per #43

PR #63, opened 2019-03-14
Fix incorrect title

mikedowler:
The key for "Install app updates from the App Store" in 10.14 is in `/Library/Preferences/com.apple.commerce.plist` which doesn't have a manifest yet.


apizz:
Thanks, @mikedowler.  Looks like com.apple.commerce only handles that one preference.  I'm going to create that manifest, but hold off on pushing it out.

I know Erik and I have talked about including preference domains within other domains to avoid having to separate these into separate payloads.

Issue #90, opened 2019-05-07
Update Smart Card payload w/ FieldTokenRemovalAction


apizz:
"Enable Screen Saver on Smart Card removal" - `FieldTokenRemovalAction`

from https://docs.jamf.com/10.12.0/jamf-pro/release-notes/What's_New.html

PR #96, opened 2019-06-12
#90 Added tokenRemovalAction to enable Screen Saver when SmartCard is…

AndrewWCarson:
… removed. Added integer range of 0-1 (default 1) for easier addition.

https://developer.apple.com/documentation/devicemanagement/smartcard?changes=latest_minor#properties

This is my first commit to this repo, so any feedback on manifest formatting would be appreciated. Thanks!

### DIFF
--- a/Manifests/ManifestsApple/com.apple.smartcard.plist
+++ b/Manifests/ManifestsApple/com.apple.smartcard.plist
@@ -192,20 +192,24 @@
 			<key>pfm_type</key>
 			<string>boolean</string>
 		</dict>
-		<dict>
+        <dict>
+			<key>pfm_default</key>
+			<integer>1</integer>
 			<key>pfm_description</key>
-			<string>If 1, enables the screen saver when the SmartCard is removed.</string>
+			<string>Enable the Screen Saver when the SmartCard is removed.</string>
 			<key>pfm_name</key>
 			<string>tokenRemovalAction</string>
 			<key>pfm_range_list</key>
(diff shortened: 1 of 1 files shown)


erikberglund:
Also, it said they use default: 1 and you can add that too for this key with the pfm_default key. Here is how I have modified this key, try it if you like locally and see if it works for you:

```
<dict>
			<key>pfm_default</key>
			<integer>1</integer>
			<key>pfm_description</key>
			<string>Enables the screen saver when the SmartCard is removed.</string>
			<key>pfm_name</key>
			<string>tokenRemovalAction</string>
			<key>pfm_range_list</key>
			<array>
				<integer>1</integer>
				<integer>0</integer>
			</array>
			<key>pfm_title</key>
			<string>Enable Screen Saver on Removal</string>
			<key>pfm_type</key>
			<string>integer</string>
			<key>pfm_type_input</key>
			<string>boolean</string>
		</dict>
```

erikberglund:
Updated the example by changing the order in the range list to work better with the bool choices.

AndrewWCarson:
Very cool that the boolean values can be configured that way. Thanks! I updated my PR with your changes.

erikberglund:
I have added this and updated the index so it is now available in the app. Do you know if this key was added in a specific version of macOS? So we could set the correct minimum os version

erikberglund:
Found it in the docs: 10.13.4. I'll add that too.

Issue #113, opened 2019-08-12
Add com.jamf.connect.shares.plist


qu1gl3s:
[https://docs.jamf.com/jamf-connect/1.0.0/sync/administrator-guide/File_Shares.html](https://docs.jamf.com/jamf-connect/1.0.0/sync/administrator-guide/File_Shares.html)

All information is located within the Jamf guide provided.

All mappings are the same as the menu.nomad.shares.plist which already exists in ProfileCreator.

kevinmcox:
Is it simply `menu.nomad.shares` pointed at a different preference domain, or are more changes besides just changing the titles and descriptions needed?

kevinmcox:
This has been added in #115.

PR #115, opened 2019-08-14
Add com.jamf.connect.shares

kevinmcox:
Initial attempt at adding `com.jamf.connect.shares` for #113.

@qu1gl3s can you test and let me know if any changes are needed, I don't have Jamf Connect here.


apizz:
@kevinmcox looks like the com.jamf.connect.shares plist is in binary.  Can we convert to xml?

kevinmcox:
@apizz Weird, not sure how I managed to do that. Should be converted now, thanks for the catch.

Issue #116, opened 2019-08-14
com.microsoft.office: add new privacy keys for 16.28


timnottom:
There are new keys for Office from version 16.28 and up. These are mentioned here: https://macadmins.software/docs/16.28-Privacy-Consent.key
More info here: https://docs.microsoft.com/nl-nl/deployoffice/privacy/mac-privacy-preferences
At time of writing the sheet with all the keys is not updated yet: https://goo.gl/UC04oZ

kevinmcox:
Thanks for filing the issue @timnottom.

kevinmcox:
Also need to update the individual Office application profiles related to the new privacy settings.

While at it, `kSubUIAppCompletedFirstRunSetup1507` needs to be updated as well in the individual app manifests.

kevinmcox:
The main Office manifest has been updated with all the new keys. I'm working on updates for all the individual application manifests now.

timnottom:
Thank you Kevin!

Issue #104, opened 2019-07-11
Add OfficeAutoSignIn to com.microsoft.office


apizz:
Per https://macadmins.slack.com/archives/C2N7MED7G/p1562834771272700

kevinmcox:
Looks like this preference already exists, but it could use description update to reflect how it influences Outlook.

I'll take care of that as part of my efforts for #116.

kevinmcox:
This is taken care of in #124.

PR #124, opened 2019-08-18
Update com.microsoft.office.plist

kevinmcox:
Major reworking to bring manifest up-to-date:
• Segmentation for easier navigation
• Added all new 16.28 keys
• Added other missing keys
• Clarified language and descriptions
• Grouped related keys and alphabetized
• Removed `HaveMergedOldPrefs` since it should not be managed

Covers the primary request for #116 (additional PRs coming) and takes care of #104 completely.

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.microsoft.office.plist
+++ b/Manifests/ManagedPreferencesApplications/com.microsoft.office.plist
@@ -6,19 +6,12 @@
 	<string>https://www.office.com</string>
 	<key>pfm_description</key>
 	<string>Microsoft Office settings</string>
-	<key>pfm_documentation_url</key>
-	<array>
-		<string>https://docs.google.com/spreadsheets/d/1ESX5td0y0OP3jdzZ-C2SItm-TUi-iA_bcHCBvaoCumw/htmlview#</string>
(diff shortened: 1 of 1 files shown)


apizz:
@kevinmcox all looks good, thanks for your efforts on this.  2 questions:

1. I don't think for the root -level `pfm_documentation_url` you can have an array of strings.  IIRC, Erik had added in links in the GUI for `pfm_documentation_url` at the preference level, but not for one or more at the root manifest level.  So if these additional links apply, I would add them to the individual preferences instead and leave the root-level one alone.
2. Were there not some preferences no longer applicable in 16.28 and later that should have a `pfm_app_max` of `16.27`?  If so, those are missing.

kevinmcox:
1. Learned something new here. Every existing `pfm_documentation_url` I've come across so far in manifests I've updated is at the root level, not preference level (which doesn't expose it in the GUI at all). I had assumed we were just listing them there as an internal reference.

2. No, those are all in the individual `com.microsoft.AppName` manifests that I'll be PR'ing separately. The new suite-wide keys here replace those individually managed keys at Office version 16.28.

kevinmcox:
I added `pfm_documentation_url` for every preference option.

PR #139, opened 2019-08-25
Add manifest and icon for Slack

homebysix:
Allows configuration of the 4.0+ autoupdate setting detailed here: https://rickheil.com/disabling-slack-updates-in-v4-0/


apizz:
Thanks for the contribution, @homebysix !  Your work here and the larger Mac Admin community is incredibly appreciated 🍻

PR #149, opened 2019-09-03
Dragonframe 4 manifest and icon

wegotoeleven:
Note, I'm having some issues with the "Suppress Update Messages" payload; for some reason, any string I add into the payload in the ProfileCreator UI doesn't stick. Any idea why?


kevinmcox:
It looks like the periods in the `pfm_name` of `updates.ignoreVersion.DRAGON4` are causing the weird behavior.

PR #150, opened 2019-09-04
com.jamf.connect.login-OpenID.plist

qu1gl3s:
Addition of the OIDCDiscoveryURL variable for ADFS environments.


qu1gl3s:
An example of this variable in use can be found here: [https://travellingtechguy.eu/jamf-connect-against-adfs-without-azure/](https://travellingtechguy.eu/jamf-connect-against-adfs-without-azure/)

PR #153, opened 2019-09-09
Compressor manifest and icon

wegotoeleven:
Compressor manifest and icon


apizz:
Will merge once we've finished #157

PR #154, opened 2019-09-09
Motion manifest and icon

wegotoeleven:


apizz:
Will merge once we've finished #157

PR #157, opened 2019-09-17
Digidna initial contributions

relgit:
Body of initial work performed at DigiDNA on the profile manifests, containing new manifests and adjustments to existing ones.

New data includes both standard and extended PFM keys, however at this time only keys supported by ProfileCreator were added.


apizz:
Thanks for all this work @relgit !  Just one item I wanted to review RE keys with spaces in the keyname, otherwise everything else seems good, and I'm sure you've looked at these changes within your app and/or ProfileCreator to verify.

@kevinmcox would you mind taking a look just to have another pair of eyes?  Nothing that stood at to me as being problematic.

kevinmcox:
Sure thing, I'll take a look this afternoon.

relgit:
Super.

Yes the spaces in the key names looked odd to me too, but that's what the docs specify ([see here](https://developer.apple.com/documentation/devicemanagement/scep/payloadcontent)). Any other key name will simply not set the desired preference in iOS.

PR #168, opened 2019-10-07
Adding BBEdit 13 licensing

n8felton:


apizz:
@n8felton Just added these prefs in our environment.  Thanks for adding them to the manifest!

PR #170, opened 2019-10-10
DigiDNA rolling contributions

relgit:
Various fixes including key-type decapitalization and standardization, text typos correction, empty fields cleanup, incorrect condition removal, and the addition of some more platform information.

Furthermore, the pre-commit and index updating scripts were modified to now work when the repository is a submodule and when ran from any directory in the project.


relgit:
@apizz nice catch, thanks!

PR #175, opened 2019-10-18
Digidna single sign on

### LINKED ISSUE #31: Add iOS Single Sign-On Account Payload
Per https://developer.apple.com/business/documentation/Configuration-Profile-Reference.pdf

relgit:
Adding com.apple.sso.
This resolves #31.


Manifests/index
```diff
@@ -1751,6 +1751,17 @@
...
+		<dict>
+			<key>hash</key>
+			<string>485e486df076b7a9c7c6a774d93086af</string>
+			<key>modified</key>
+			<date>2019-10-18T16:49:24Z</date>
+			<key>path</key>
+			<string>Manifests/ManifestsApple/com.apple.sso.plist</string>
+			<key>version</key>
+			<integer>1</integer>
+		</dict>
```
apizz: In the future, would you mind not updating the manifest indexes as part of your commits?  That way if we're merging multiple or different PRs first we avoid a commit conflict.  Whomever is doing the PR merges should be the one to update the indexes.

relgit: You got it @apizz.

PR #179, opened 2019-11-07
DigiDNA default values

relgit:
This PR adds missing default values to (mostly-) boolean keys in Apple domains

It also contains a few corrections to preexisting default values which by chance were found to be wrong, but this is not an exhaustive verification of all of them.


apizz:
@relgit only thing I think that's needed is the `pfm_last_modified` dates to be updated.  Since we don't have new preferences here, no need to bump the version.

PR #180, opened 2019-11-11
DigiDNA Universal Access improvements

relgit:
Domain com.apple.universalaccess brought closer in line with the documentation.

The properties _contrast_, _mouseDriverCursorSize_, and _mouseDriverInitialDelay_ were left with type 'real' as the documentation contradicts practice.


relgit:
That must have slipped in by mistake, thanks for noticing. If I revert this commit then that would undo other commits on the master, so I'm just going to cancel this PR and and make a new one.

PR #181, opened 2019-11-13
Create com.brave.Browser.manifest

staze:
Just adding manifest for Brave Browser.


apizz:
Thanks for this first effort @staze.  While you have the preferences themselves, there's a lot of things missing here for the manifest itself in order to work in ProfileCreator.  Please have a look at https://github.com/ProfileCreator/ProfileManifests/wiki/Getting-Started.  You may find reviewing some of the other manifests helpful, as well.

That being said, since Brave pulls from the open-source Chromium, these preferences are effectively duplicated by Google Chrome.  As a result, my thought here is just to duplicate what's configured in the Chrome manifest albeit change the domain and title to allow its use for Brave, rather than move forward with this PR.

Let me know your thoughts.

staze:
Hi @apizz,

Fine by me. =) I honestly came here looking for the profile, didn't find it. asked on their forum, then actually LOOKED in the app and found it. So yeah, guessing just changing the domain and we're golden. 

Thanks for getting back to me. feel free to cancel PR.

apizz:
Closing per https://github.com/ProfileCreator/ProfileManifests/commit/6195d247805d55e95709bab3885d0568fe9bf570

Issue #178, opened 2019-10-30
Add Privileges.app payload


apizz:
ex. https://github.com/SAP/macOS-enterprise-privileges/blob/master/application_management/example_profile/Privileges.mobileconfig

Per https://macadmins.slack.com/archives/C2N7MED7G/p1572457836094300

apizz:
Completed

PR #186, opened 2019-12-02
SAP Privileges App  Manifest

HowardGMac:
Adding a manifest for the SAP Privileges app that as of version 1.0.5 has configuration profile keys.


apizz:
@HowardGMac made some additions, but looks good otherwise.  Thanks!

Per #178

PR #194, opened 2019-12-09
DigiDNA MCX Improvements

relgit:
Improvements to the three MCX subdomains, Energy Saver, Time Server, and Mobile Accounts.

These changes focus on alignment to the documentation. Missing keys were introduced and a number of undocumented ones were removed. Keys were removed where no documentation for them could be found on either Google or the MacAdmins Slack.

In addition, there are a few presentational tweaks.

### DIFF
--- a/Manifests/ManifestsApple/com.apple.MCX-EnergySaver.plist
+++ b/Manifests/ManifestsApple/com.apple.MCX-EnergySaver.plist
@@ -8,12 +8,14 @@
 	<string>com.apple.MCX</string>
 	<key>pfm_subdomain</key>
 	<string>EnergySaver</string>
+	<key>pfm_documentation_url</key>
+	<string>https://developer.apple.com/documentation/devicemanagement/energysaver</string>
 	<key>pfm_format_version</key>
 	<integer>1</integer>
 	<key>pfm_interaction</key>
 	<string>combined</string>
 	<key>pfm_last_modified</key>
-	<date>2018-07-18T08:58:49Z</date>
+	<date>2019-12-09T17:40:13Z</date>
 	<key>pfm_platforms</key>
 	<array>
 		<string>macOS</string>
@@ -119,6 +121,7 @@
 			<string>PFC_SegmentedControl_0</string>
 			<key>pfm_range_list_titles</key>
 			<array>
+				<string>General</string>
 				<string>Desktop</string>
 				<string>Portable</string>
 				<string>Schedule</string>
@@ -127,6 +130,11 @@
 			<string>always</string>
 			<key>pfm_segments</key>
 			<dict>
+				<key>General</key>
+				<array>
+					<string>DestroyFVKeyOnStandby</string>
+					<string>SleepDisabled</string>
+				</array>
 				<key>Desktop</key>
 				<array>
 					<string>com.apple.EnergySaver.desktop.ACPower</string>
@@ -155,22 +163,8 @@
 			<key>pfm_subkeys</key>
 			<array>
 				<dict>
-					<key>pfm_default</key>
-					<integer>-1</integer>
 					<key>pfm_description</key>
-					<string></string>
-					<key>pfm_hidden</key>
-					<string>all</string>
-					<key>pfm_name</key>
-					<string>com.apple.EnergySaver.desktop.ACPower-ProfileNumber</string>
-					<key>pfm_title</key>
-					<string>Desktop Energy Saver Profile</string>
-					<key>pfm_type</key>
-					<string>integer</string>
-				</dict>
-				<dict>
-					<key>pfm_description</key>
-					<string></string>
+					<string>Minutes of inactivity before the display will be turned off, in an integer value where 0 means never</string>
 					<key>pfm_name</key>
 					<string>Display Sleep Timer</string>
 					<key>pfm_range_list</key>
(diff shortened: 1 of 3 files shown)


Manifests/ManifestsApple/com.apple.MCX-EnergySaver.plist
```diff
@@ -155,24 +163,58 @@
...
+						<string>10 Minutes</string>
+						<string>20 Minutes</string>
+						<string>30 Minutes</string>
+						<string>45 Minutes</string>
+						<string>1 Hour</string>
+						<string>1 Hour 15 Minutes</string>
+						<string>1 Hour 30 Minutes</string>
+						<string>1 Hour 45 Minutes</string>
+						<string>2 Hours</string>
+					</array>
```
apizz: I completely understand the desire to be more in line with the docs, but I wonder if by adding these defined times if we're not unnecessarily limiting what's possible for admins to configure.  For someone who _might_ want to have the display go to sleep after 25 minutes, or some other custom option within the acceptable range of options, I'm hesitant to have locked in options like this.

apizz: I may have spoken too soon ... it looks like these were in the profile to begin with.

Manifests/ManifestsApple/com.apple.MCX-MobileAccounts.plist
```diff
@@ -153,12 +157,72 @@ The payload organization for a payload need not match the payload organization i
...
+			<key>pfm_description</key>
+			<string></string>
+			<key>pfm_description_reference</key>
+			<string>If true, allows the user to stop the prompts about mobile account creation every time the user logs in. This key is only valid if com.apple.cachedaccounts.WarnOnCreate is set to true.</string>
+			<key>pfm_name</key>
+			<string>cachedaccounts.WarnOnCreate.allowNever</string>
+			<key>pfm_title</key>
+			<string>Allow Users to Stop Mobile Account Creation Prompts</string>
+			<key>pfm_type</key>
+			<string>boolean</string>
```
apizz: Suggesting the addition of the following code.  Reason being is the only time this key should be used is in conjunction with `com.apple.cachedaccounts.WarnOnCreate` when it's set to `true`.  If someone doesn't want users to be prompted about Mobile Account creation, then `com.apple.cachedaccounts.WarnOnCreate` set to false and therefore exclude`com.apple.cached.accounts.allowNever` if added.

```
<key>pfm_exclude</key>
<array>
	<dict>
		<array>
			<dict>
				<key>pfm_present</key>
				<false/>
				<key>pfm_range_list</key>
				<array>
					<false/>
				</array>
				<key>pfm_target</key>
				<string>com.apple.cachedaccounts.WarnOnCreate</string>
			</dict>
		</array>
	</dict>
</array>
```

apizz: FYI, that code ^ has been tested successfully.

Manifests/ManifestsApple/com.apple.MCX-TimeServer.plist
```diff
@@ -145,33 +149,27 @@ The payload organization for a payload need not match the payload organization i
...
-Availability: Available in macOS 10.12.4 and later.</string>
+			<string>The ntp server to connect to. Availability: Available in macOS 10.12.4 and later.</string>
 			<key>pfm_name</key>
 			<string>timeServer</string>
 			<key>pfm_title</key>
 			<string>Time Server</string>
 			<key>pfm_type</key>
 			<string>string</string>
-			<key>pfm_macos_min</key>
-			<string>10.12.4</string>
```
apizz: Time Server payload in Apple config profile reference (https://developer.apple.com/business/documentation/Configuration-Profile-Reference.pdf) includes this info, so would like to keep the macOS min data

Manifests/ManifestsApple/com.apple.MCX-EnergySaver.plist
relgit: I'm unresolving this as it may be worth a bit more discussion.

See, I was thinking along the same lines as you - I do prefer giving admins power and granularity, and in fact in IPE all similarly ranged fields will be reverted back to free-form integer ones.

The reason that I added them here was indeed to match the already existing fields that you mentioned. But, if we're both in agreement over this, perhaps it would be better to remove all range lists from timer properties in this domain.

What do you think?

apizz: I'd vote to remove the concrete range options.

relgit: I removed all range lists on timer properties in commit 046487b below.

Manifests/ManifestsApple/com.apple.MCX-MobileAccounts.plist
relgit: Seems like a good idea. Added via commit `dca5fff`.

Manifests/ManifestsApple/com.apple.MCX-TimeServer.plist
relgit: I think this was a per-property leftover from the old undivided MCX manifest before it was split into subdomains. Since both keys in this subdomain have the same requirement, we moved it to the root of the manifest on lines 26-27.

apizz:
I'll take a more comprehensive look at the resulting changes soon and get this merged.  On holiday so going to try and disconnect :)

relgit:
Of course, enjoy your holiday and thanks for the help 😀

PR #196, opened 2019-12-27
DigiDNA Dec 2019 Contributions

relgit:
Various tweaks:

- Removed the file-type key from a property it doesn't belong to on the _Fonts_ domain. Note that it is still present on the correct property.
- Added a default value to a top-level property.
- Narrowed allowed file type according to the docs on the _WebClips_ domain.
- On the _WiFi_ domain: new keys added, Interface pseudo-key `Interface` was migrated to `PFC_InterfaceSelector` since Apple introduced a native one by the same name, conditional added to `SSID_STR`, and texts were tweaked.

### DIFF
--- a/Manifests/ManifestsApple/Configuration.plist
+++ b/Manifests/ManifestsApple/Configuration.plist
@@ -11,7 +11,7 @@
 	<key>pfm_interaction</key>
 	<string>exclusive</string>
 	<key>pfm_last_modified</key>
-	<date>2019-10-10T09:37:17Z</date>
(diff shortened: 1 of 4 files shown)


relgit:
It's not in the docs, it's just a ghost property to help hide some of the clutter of the manifest.

If you look, you can see that it has a `pfm_excluded` key set to true on it.

We can't use the original name (`Interface`) anymore because Apple now has a key with the same name in the payload. I renamed it to `PFC_...` based on other ghost keys I found in the repo.

apizz:
All looks good to me.  Thanks!

PR #197, opened 2019-12-30
DigiDNA tvOS Restrictions update

relgit:
A quick addition of app white-and-black list properties.


apizz:
Afk but think there's a key to specify that supervision on a pref key is required.  Can that be added?

PR #218, opened 2020-01-29
DigiDNA Software Updates

relgit:
The properties `forceDelayedSoftwareUpdates` and `enforcedSoftwareUpdateDelay` on the Restrictions domain were locked in a dependency reference cycle. This prevented the former property from being written to a profile without the latter.

According to the Apple docs a profile containing `forceDelayedSoftwareUpdates` by itself is valid. This was also confirmed with a test on a supervised iOS device.

We resolved the reference cycle and in addition streamlined the remaining conditions. Furthermore, the minimum platform version was noted for tvOS.


relgit:
Closed while running housekeeping; Reopened.

apizz:
Your latest merge has all of the recent commits, not just the ones you changed

relgit:
Thanks for taking a look at this @apizz.

Yes, that merge is the result of resolving a conflict between our branch and `master`, and it will happen again if I reproduce this PR because the conflict will still have to be resolved.

You can proceed with confidence though since the merged commits originated in the ProfileCreator/ProfileManifests `master` and will not introduce any further changes beyond my original two commits. The diff between the branches illustrates this best- https://github.com/ProfileCreator/ProfileManifests/compare/master...DigiDNA:digidna_software_updates.

apizz:
I learned something.  Thanks.

PR #225, opened 2020-02-06
Adding missing key and update another to Google Chrome 

TK5-Tim:
I needed AllowSyncXHRInPageDismissal in my Google Chrome Profile so I added it. Also I saw that for AllowPopupsDuringPageUnload the deprecation version changed.

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.google.Chrome.plist
+++ b/Manifests/ManagedPreferencesApplications/com.google.Chrome.plist
@@ -15,7 +15,7 @@
 	<key>pfm_interaction</key>
 	<string>combined</string>
 	<key>pfm_last_modified</key>
-	<date>2019-12-10T17:48:00Z</date>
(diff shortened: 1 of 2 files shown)


apizz:
Thanks @TK5-Tim.  Can you please remove the index update?  There are occasions when multiple PRs come in and one is merged before the other and they both update the index then we have a problem.  Whenever I commit a merge is when I update the index and want to keep things consistent.

TK5-Tim:
@apizz Sure I can toally understand that! Should I keep the pfm_version update in there or should I remove it too?

apizz:
@TK5-Tim anything you've done in Chrome looks OK.  If you just revert your most recent commit just add a new commit with Chrome's pfm_version and date modified as you had originally.  Much appreciated

TK5-Tim:
@apizz The revert did not work but i took out all the changes I did to the index file so now it is the same as in ProfileManifests/master.

Issue #213, opened 2020-01-20
Fix IsHotspot pref in Wi-Fi payload


apizz:
When added manually, displays a blank area where an exclusion should likely exist. Happens in both light and dark modes.

![Screen Shot 2020-01-20 at 12 37 54 PM](https://user-images.githubusercontent.com/11789931/72747405-8c120580-3b82-11ea-871e-42f088cc9d11.png)

 When 'Network Interface' is set to 'Legacy Hotspot' this preference is automatically added and shows up just fine.

![Screen Shot 2020-01-20 at 12 41 49 PM](https://user-images.githubusercontent.com/11789931/72747398-874d5180-3b82-11ea-9ad0-625a6583a92a.png)

PR #234, opened 2020-02-27
Update several aspects of the Wi-Fi payload

apizz:
- Add pfm_description_reference & pfm_note for AcceptEAPTypes
- Add missing TLSCertificateIsRequired preference
- Add pfm_default and pfm_description_reference for TLSMaximumVersion and TLSMinimumVersion, remove blank pfm_description
- Remove instances of 2 pfm_target_conditions dictionaries within pfm_conditionals, when multiple options should be listed under 1.
- Add pfm_exclude for SystemModeUseOpenDirectoryCredentials & SystemModeCredentialsSource, as docs indicate only 1 is supposed to be used. Use pfm_note to reinforce the fact that only one of these keys are to be used in a profile. Also add pfm_description_reference.
- Add pfm_range_list_titles array for SystemModeCredentialsSource to have the more friendly 'Active Directory' option listed
- Remove Interface preference since the options are only related to ethernet connections, as these are being broken out into separate manifests (upcoming PR)
- Add pfm_present key with `false` value for pfm_exclude dictionaries when the pfm_target is EAPClientConfiguration.AcceptEAPTypes
- Bump pfm_version and modified date

### DIFF
--- a/Manifests/ManifestsApple/com.apple.wifi.managed.plist
+++ b/Manifests/ManifestsApple/com.apple.wifi.managed.plist
@@ -13,7 +13,7 @@
 	<key>pfm_interaction</key>
 	<string>combined</string>
 	<key>pfm_last_modified</key>
-	<date>2020-01-17T11:58:00Z</date>
+	<date>2020-02-26T15:48:00Z</date>
 	<key>pfm_platforms</key>
 	<array>
 		<string>iOS</string>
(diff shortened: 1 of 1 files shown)


apizz:
This does not address #213 ATM ... but it will ...

relgit:
Many of the changes in this PR have over time been implemented. Changes that haven't should be merged but yet-to-be-submitted work on this manifest would make that impractical. Instead, I implemented these changes on fork branches that are going to be submitted soon (see some references above).

The only changes that are not already implemented or that I haven't included are the target-conditions related ones, which I don't believe are necessary. I did test the current implementation on both iPE and PC and can confirm that it works as intended.

If no one objects, I'm going to close the PR.

PR #236, opened 2020-03-11
Update for Privileges.app 1.5.0 keys.

peetinc:
This is the first pull request I've ever done, so please bear with me if I screwed it or some protocol up. I just updated this for the new keys in 1.5.0 and above as documented here: https://github.com/SAP/macOS-enterprise-privileges/tree/master/application_management

Thanks.Peet


apizz:
Merged into #239

PR #258, opened 2020-04-15
Add FileVaultOptions Payload

apizz:
Creates a separate FileVault Options payload, which is configured as a subdomain of the larger `com.apple.MCX` payload.

- Add `dontAllowFDEDisable` & `dontAllowFDEEnable` preferences
- Move `DestroyFVKeyOnStandby` from Energy Saver manifest to FileVault Options.
  - Added `pfm_documentation_url` to energy saver documentation and `pfm_note` to indicate where the pref lives


apizz:
Change checked & OK'd by DigiDNA to be addressed in iMazing Profile Editor

PR #260, opened 2020-04-16
DigiDNA Ethernet tweak

relgit:
I noticed that the Ethernet manifests were set to allow multiple payloads so here's a correction of that.


apizz:
Thank you for catching this!

PR #270, opened 2020-04-30
Web Filter manifest

apizz:
This was originally to address the fact that no manifest existed, but it did it just had the wrong name and not all the right preferences.

I'm a little confused by the 'enable' preferences, as per Apple's documentation these are indeed `boolean`, but looking at the same manifest in the latest version of Apple Configurator 2 these are listed as being `integer` with `0` = disabled and `1` = enabled.

@relgit can you confirm?


relgit:
@apizz I did some more work on this PR, mainly on conditions. Please check to see that everything looks good for you, and if so I believe we're ready to merge. Thanks.

relgit:
Also, per your question about the _Integer_ vs _Boolean_ discrepancy, I believe that it ultimately doesn't matter, because both types are represented under the hood as `NSNumber` objects (_Real_ numbers too btw).

So, while Booleans are marked with `<false/>` and `<true/>` XML elements when serialized to a Plist, when loaded by macOS or iOS, they become `NSNumber` objects with the values `0` and `1` respectively. If ever the system needs a boolean value for a certain key it calls the `boolValue` method on that loaded object and it then gets such a value, no matter if the data was de-serialized from either Boolean or Integer.

PR #280, opened 2020-05-20
Add a few default values to placeholders in com.trusourcelabs.NoMAD.plist

joncrain:
Adding the default values of TitleSignIn and MessageLocalSync to the placeholder text.

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.trusourcelabs.NoMAD.plist
+++ b/Manifests/ManagedPreferencesApplications/com.trusourcelabs.NoMAD.plist
@@ -13,7 +13,7 @@
 	<key>pfm_format_version</key>
 	<integer>1</integer>
 	<key>pfm_last_modified</key>
-	<date>2020-05-14T11:47:11Z</date>
+	<date>2020-05-20T13:35:11Z</date>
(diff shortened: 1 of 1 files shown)


Manifests/ManagedPreferencesApplications/com.trusourcelabs.NoMAD.plist
```diff
@@ -1434,7 +1434,7 @@ Line two</string>
 			<key>pfm_type</key>
 			<string>string</string>
 			<key>pfm_value_placeholder</key>
-			<string>Please provide your local password.</string>
+			<string>Your network and local passwords are not the same. Please enter the password for your Mac.</string>
```
apizz, line 1437: Isn't the assumption about MessageLocalSync that the *local password* does not match the network password and therefore they need to enter their *network* password to resync with the local account?

If that's the case then to me the wording for the placeholder should be different.

apizz: If I had only read the documentation ...

apizz:
@joncrain make sure to at the very least bump the modified date in the manifest.  Thank you for the contribution.

PR #308, opened 2020-06-25
DigiDNA Brave Browser Conversion

### LINKED ISSUE #297: Replace mentions of Chrome in the Brave manifest

**Name of App / Payload**: *ex. Firefox OR org.mozilla.firefox*
com.brave.Browser

**Preference key(s) needing fixing**: *ex. `AppUpdateURL`*
DefaultBrowserSettingEnabled (for example)

**Description of problem(s) / error(s) with listed key(s)**: *ex. When `X` preference is configured, it does not *
Preference key titles and description mention Chrome.

**Description of expected behavior**:
They should mention Brave instead.

**Applicable Documentation URL(s)**: *If applicable ...*
https://support.brave.com/hc/en-us/articles/360039248271-Group-Policy

**How are you testing / deploying your profile?**: *ex. manual profile install, MDM, etc.*
IPE

relgit:
Script and data set to automatically generate the `com.brave.Browser` manifest from `com.google.Chrome` every time that the latter updates.

The script can be run from any subdirectory in the repository, takes no arguments, and requires Python3.

Future Brave-specific subkeys (currently there's only one) should be added to `Resources/ChromeToBrave/com.brave.Browser-specific.plist` before running the script.

Includes replacement of all Chrome mentions where appropriate which resolves #297.

### DIFF
--- a/Resources/ChromeToBrave/ChromeToBrave.py
+++ b/Resources/ChromeToBrave/ChromeToBrave.py
@@ -1,0 +1,122 @@
+#!/usr/local/bin/python3
+# -*- coding: utf-8 -*-
+
+import os
+import subprocess
+import sys
+import plistlib
+
+chrome_manifest_domain = "com.google.Chrome"
+brave_manifest_domain = "com.brave.Browser"
+manifests_subfolder = os.path.join( "Manifests", "ManagedPreferencesApplications" )
+
+def main():
+    # Get path to repository root
+    popen_args = [ "git", "rev-parse", "--show-toplevel" ]
+    repository_root_process: subprocess.Popen = subprocess.Popen( popen_args, stdout = subprocess.PIPE, stderr = subprocess.STDOUT )
+    ( stdout, stderr ) = repository_root_process.communicate()
+
+    if repository_root_process.returncode != 0:
+        print( "Getting repo root path errored out:\n" + stdout )
+        exit( 1 )
+
+    repository_root_path = stdout.decode( sys.stdout.encoding ).strip()
+    chrome_manifest_path = os.path.join( repository_root_path, manifests_subfolder, chrome_manifest_domain + ".plist" )
+
+    # Load the Chrome manifest
+    try:
+        chrome_manifest_file = open( chrome_manifest_path, 'rb' )
+        manifest = plistlib.load( chrome_manifest_file )
+    except Exception as error:
+        print( "Could not load " + chrome_manifest_path )
+        print( "Exception message: " + str( error ) )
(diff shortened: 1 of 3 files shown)


apizz:
this is wonderful!  Will review when I'm able

Resources/ChromeToBrave/ChromeToBrave.py
```diff
@@ -0,0 +1,136 @@
...
+
+import os
+import subprocess
+import sys
+import plistlib
+import re
+
+chrome_manifest_domain = "com.google.Chrome"
+brave_manifest_domain = "com.brave.Browser"
+manifests_subfolder = os.path.join( "Manifests", "ManagedPreferencesApplications" )
```
apizz, line 12: This assumes that the current working directory is the ProfileManifests folder.  If that's not the case, I'd love some output to indicate this fact.

Resources/ChromeToBrave/ChromeToBrave.py
```diff
@@ -0,0 +1,136 @@
...
+                    if "pfm_name" in specific_key:
+                        misc_segment.append( specific_key[ "pfm_name" ] )
+
+    # Write-out
+    brave_manifest_path = os.path.join( repository_root_path, manifests_subfolder, brave_manifest_domain + ".plist" )
+    brave_manifest_file = open( brave_manifest_path, 'wb' )
+    plistlib.dump( manifest, brave_manifest_file )
+    brave_manifest_file.close()
+
+    print( "Done." )
```
apizz, line 89: I'd like something a little more descriptive.  Something like "Brave manifest successfully created at <absolute_path>"

relgit:
These were easy changes to make. Check it out now @apizz  and let me know.

apizz:
Thanks @relgit.  The only other issue I see now that I didn't before is that the `pfm_last_modified` date appears to be taken from the Chrome manifest, rather than reflect the date and time `ChromeToBrave.py` is run.  After running, this is what I see compared to your Brave manifest changes w/in GitHub Desktop.  Subsequent runs don't alter this.

```
	<key>pfm_interaction</key>
	<string>combined</string>
	<key>pfm_last_modified</key>
-	<date>2020-06-25T13:50:27Z</date>
+	<date>2020-04-14T09:48:00Z</date>
	<key>pfm_platforms</key>
	<array>
		<string>macOS</string>
```

Thanks again for your work on this.

relgit:
That one was also easy to add, though at this point @apizz you've used all of your three wishes (even though you didn't even know you had them 😎 ).

PR #324, opened 2020-07-15
1Password

kevinmcox:
Added auto-update related preferences for 1Password.


kevinmcox:
_This setting has no effect unless checking for updates is disabled. Meaning you cannot force automatic checking but disable automatic installation._

You're not wrong, it appears to be a bug with 1Password.

The limitation is that you cannot force enable update checking (`CheckForSoftwareUpdatesEnabled`) and disable automatic installation (`AutoInstallSoftwareUpdatesEnabled`) which I could see as a useful combination.

If `AutoInstallSoftwareUpdatesEnabled` is `true`, then whatever you configure for `AutoInstallSoftwareUpdatesEnabled` is ignored and the end user can change the setting at will.

However if `AutoInstallSoftwareUpdatesEnabled` is `false` then 1Password will enforce either value for `AutoInstallSoftwareUpdatesEnabled` and disable it in the GUI. Obviously a setting of true wouldn't actually do anything, but it shows the application is aware of the option and the key in the profile is correct.

So it isn't a problem with PC or the exported profile, but a problem with 1Password. That is what I was trying to convey in the note. I'm hoping I can get a bug report filed and have them fix it in a future update.

kevinmcox:
I guess we could just remove the `AutoInstallSoftwareUpdatesEnabled` option completely for now since it is effectively useless. Then add it back in when they fix it.

I was just going for completeness but it might be best to just avoid the confusion.

apizz:
I'd vote to include in manifest since the work has been done, but comment it out until it's properly addressed

kevinmcox:
Sounds good, that change is made.

apizz:
The only thing I added was a pfm_exclude key in prep for this behaving as we expect it to which requires checks to be both present & enabled

MikeSharkey1:
Could someone include something to enable/enforce beta installs and updates? It's useful as the beta version is required for the integration with the 1Password X Chrome extension.

kevinmcox:
Any chance you know the keys to manage that setting?

MikeSharkey1:
@kevinmcox Unfortunately not.

apizz:
@MikeSharkey1 If you find additional preferences to add, just create a new issue with links and we can get them added.

kevinmcox:
@MikeSharkey1 I saw the the latest update to the iOS app included new additions to items that can be managed with a profile. So it'll just be a matter of finding out if the settings you want can be managed or not in the desktop app, and if not requesting that 1Password make them manageable.

PR #325, opened 2020-07-22
Add S3 Endpoint for S3-Auth Plugin

peetinc:
Wasabi (and I'm assuming other S3 compatible hosts) needs to have the S3Endpoint key defined to function. Without the key S3-Auth defaults to s3.amazonaws.com.

`S3_ENDPOINT = pref('S3Endpoint') or 's3.amazonaws.com`


apizz:
Looks good. Thanks for the contribution! Will merge soon.

PR #327, opened 2020-07-24
Add ShareKit manifest

apizz:
Per discussion in #profilecreator Slack channel (https://macadmins.slack.com/archives/C2N7MED7G/p1595515885293400) adds the long-deprecated ShareKit manifest.  Unsure if this is actually respected in later OSs (10.13+), but payload is still included in Jamf's profiles (albeit it's been broken for a while, video: https://macadmins.slack.com/archives/C2N7MED7G/p1595516156295000).  At the very least, want to give admins the opportunity to suppress these share options if we're able to.

### DIFF
--- a/Manifests/ManifestsApple/com.apple.ShareKitHelper.plist
+++ b/Manifests/ManifestsApple/com.apple.ShareKitHelper.plist
@@ -1,0 +1,274 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>pfm_description</key>
+	<string>Configures available Share menu options</string>
+	<key>pfm_documentation_url</key>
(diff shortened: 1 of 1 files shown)


apizz:
Confirmed on Late 2012 MBP that an empty `SHKAllowedShareServices` array that all share options are gone!

bp88:
> I don't have access to a Jamf instance anymore.  Would you be able to share the pref key for auto enabling additional sharing services?  It's not in the docs
It's hard to tell if this is a bug or not but the Jamf interface will spit out the following if you want to automatically allow everything going forward::

```
			<key>SHKDeniedShareServices</key>
			<array/>
```

I take it to mean that this is saying there should be no Denied Share Services which should be interpreted as the OS should not block anything.

This payload is only supported from 10.9 to 10.12 as noted as noted earlier and has since been deprecated which means it may stop working in a future OS: https://developer.apple.com/documentation/devicemanagement/sharekit

apizz:
That makes sense to me. I did test the profile on a 10.14.6 Mac and it worked as expected. Was just curious if the checkbox I saw in Jamf was a separate key

PR #334, opened 2020-08-05
Removal of erroneous RecentWhatsNewPanelVersion key

wegotoeleven:
The current version of this manifest has two `RecentWhatsNewPanelVersion` keys. I've removed the dodgy one.


apizz:
Good catch! Thanks.

PR #335, opened 2020-08-05
Handbrake config and icon

wegotoeleven:
New config for Handbrake :)


apizz:
Thanks for this @wegotoeleven !  I added some additional Sparkle settings.  Shouldn't be problematic to set if the app doesn't support them.

PR #341, opened 2020-08-06
McNeel Rhinoceros

wegotoeleven:
Icon and manifest!


apizz:
I don't see `SUSkippedVersion` on the sparkle documentation page.  Is that one just specific to Rhino, or something you've sleuthed from messing with the app preferences?

wegotoeleven:
Definitely by sleuthing :) Also, this: https://github.com/sparkle-project/Sparkle/issues/1142

apizz:
How would you intend an admin to use that preference?  It could just be me, but I can't imagine using that preference.  I'd either let the app keep itself updated or disable it and handle installs via other mechanisms.  It comes across as situational to me - not something you would implement in a profile long term.

wegotoeleven:
Thing is, this needs to be set, as this is the key that suppresses the update notification for a specific version. The version number doesn't have to apply to a specific version though, as I just use `9.9.9.9` to supress everything from the current version and above.

Without setting this, the update prompt is still displayed at launch.

PR #344, opened 2020-08-06
Update ChromeToBrave.py script to include Sparkle update keys

### LINKED ISSUE #342: Add Sparkle preferences to Brave manifest

**Name of App / Payload**: Brave

**Preference key(s) to be added / updated**: Sparkle prefs added in #335 

Will need to update the ChromeToBrave.py script to include these values, since Google uses its own update framework, not Sparkle.

Needs:
- Add an 'Updates' tab to segments
- Add the preference key names to an 'Updates' array in segments
- Add the preferences to the manifest

apizz:
Still a work in progress, and need feedback given my limited Python skills ...

Previously, we just had an array of dictionaries with preferences that were all appended to the bottom of the manifest and appended to the 'Misc.' menu segment.  Given the addition of other preferences that we do not want to have fall under 'Misc.', my thought is in order to better handle additional Brave-specific preferences moving forward is two-fold:

1. Restructure the com.brave.Browser-specific.plist to instead be a dictionary where the keys match the segment name.
2. The segment keys are arrays with the preference keys as contained dictionaries.

Right now, the segment key is being written as needed, but the preferences themselves are not (below).  On that front, @relgit I could use some help 😄  

Screen Shot 2020-08-06 at 7 07 19 PM
Screen Shot 2020-08-06 at 7 07 27 PM

Closes #342


apizz:
@relgit this is now working as expected!  Proved to be more challenging to wrap my brain around, but it works.

The one thing this doesn't do at the moment but would like it to do is to not add a preference in the domain specific plist if it already exists in the Chrome manifest.  So, if a given preference (pfm_name) already exists in `segments`, the corresponding pfm_name should not be written again to the segments array and the preference dictionary should not be written for that item.

apizz:
@relgit not trying to bother if you're busy, just proud that I was able to figure this out on my own :). Whenever you have time to review so I can merge this, 👍

relgit:
Thanks for waiting @apizz. Looking good.

I'd argue that if a preference key is both in the Chrome and the Brave-specific manifests, then the Brave-specific should prevail. This would allow us to add rudimentary modifications to existing Chrome keys if necessary.

So basically as far as I'm concerned you're good to move ahead and merge.

apizz:
Awesome, thank you @relgit ! I'm starting to work on a ChromeToEdge PR as well, but is a bit different since there are a number of chrome prefs edge doesn't use as well as prefs edge has that chrome doesn't. Hoping to get your help on that once I PR it.

apizz:
Ooof I haven't looked at this in a while.  I'm going to get this committed since it is in a working state.

PR #352, opened 2020-08-15
Add mid-2020 iMac to touristd manifest

apizz:


apizz:
Oops, need to fix the `pfm_name` before merging ...

apizz:
URL does not appear to be live yet, but we'll assume it's correct ...

PR #358, opened 2020-08-23
Add VPN Manifest

### LINKED ISSUE #32: Add VPN Payload (com.apple.vpn.managed)
Per https://developer.apple.com/business/documentation/Configuration-Profile-Reference.pdf

apizz:
NOT YET READY, just want to have a formal place to document changes as we go ...

Closes #32


apizz:
First issue I've run into:

pfm_name `PPP` is a dictionary, but which the pfm_conditionals and pfm_exclude differs from the one Apple Configurator 2 has.

Observation: When `PPTP` is selected as the VPN type, no preferences are added.  Additionally, when the PPP dictionary is added manually, which is listed as "PPP Settings for L2TP and PPTP VPN types", it incorrectly displays a pfm_exclude error message.

Screen Shot 2020-08-23 at 7 07 47 PM

Dummy profile in configurator settings & resulting profile w/ PPTP selected:

Screen Shot 2020-08-23 at 7 21 40 PM

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>PayloadContent</key>
	<array>
		<dict>
			<key>IPv4</key>
			<dict>
				<key>OverridePrimary</key>
				<integer>0</integer>
			</dict>
			<key>PPP</key>
			<dict>
				<key>AuthName</key>
				<string>test2</string>
				<key>AuthPassword</key>
				<string>test3</string>
				<key>CCPEnabled</key>
				<integer>1</integer>
				<key>CCPMPPE128Enabled</key>
				<integer>1</integer>
				<key>CCPMPPE40Enabled</key>
				<integer>1</integer>
				<key>CommRemoteAddress</key>
				<string>test1</string>
			</dict>
			<key>PayloadDescription</key>
			<string>Configures VPN settings</string>
			<key>PayloadDisplayName</key>
			<string>VPN</string>
			<key>PayloadIdentifier</key>
			<string>com.apple.vpn.managed.46F60821-7162-4259-8338-8D6A7FF2C848</string>
			<key>PayloadType</key>
			<string>com.apple.vpn.managed</string>
			<key>PayloadUUID</key>
			<string>46F60821-7162-4259-8338-8D6A7FF2C848</string>
			<key>PayloadVersion</key>
			<integer>1</integer>
			<key>Proxies</key>
			<dict>
				<key>HTTPEnable</key>
				<integer>0</integer>
				<key>HTTPSEnable</key>
				<integer>0</integer>
			</dict>
			<key>UserDefinedName</key>
			<string>test</string>
			<key>VPNType</key>
			<string>PPTP</string>
		</dict>
	</array>
	<key>PayloadDisplayName</key>
	<string>Untitled</string>
	<key>PayloadIdentifier</key>
	<string>APs-iMac.EA8AA2DF-C72C-4F5F-A123-49645C2CA2A0</string>
	<key>PayloadRemovalDisallowed</key>
	<false/>
	<key>PayloadType</key>
	<string>Configuration</string>
	<key>PayloadUUID</key>
	<string>ECE1628E-0ED6-4BDB-8C4E-05E41E0B579C</string>
	<key>PayloadVersion</key>
	<integer>1</integer>
</dict>
</plist>
```

I'm able to get the exclusion message to not appear when I comment out the `LT2P` option and ONLY leave `PPTP` listed.  The reverse is true when I select `LT2P` as the VPN type and comment out the `PPTP` option in the `pfm_exclude`.  What this tells me is that using `pfm_range_list` and `pfm_n_range_list` assumes AND, meaning both options must be true, not OR aka one or the other.

The only way I'm able to get around this is by instead using `pfm_n_contains_any`.  So the pfm_exclude would be:
```
<key>pfm_exclude</key>
<array>
	<dict>
		<array>
			<dict>
				<key>pfm_n_contains_any</key>
				<array>
					<string>PPTP</string>
					<string>L2TP</string>
				</array>
				<key>pfm_target</key>
				<string>VPNType</string>
			</dict>
		</array>
	</dict>
</array>
```

So for any additional pfm_excludes which have multiple options we'd likely have to do something similar.

While the original VPN manifest does _not_ have a `pfm_conditionals` for the PPP dictionary, I can get Server (`CommRemoteAddress`) to automatically add with it added.

```
<key>pfm_conditionals</key>
<array>
	<dict>
		<array>
			<dict>
				<key>pfm_contains_any</key>
				<array>
					<string>PPTP</string>
					<string>L2TP</string>
				</array>
				<key>pfm_target</key>
				<string>VPNType</string>
			</dict>
		</array>
	</dict>
</array>
```

Screen Shot 2020-08-23 at 7 45 51 PM
  
Screen Shot 2020-08-23 at 7 31 13 PM

As a result, of this, I'm wondering if for the moment we want to maintain two separate versions of this manifest for comparison & testing purposes:

1) One that adheres to Apple's original VPN as closely as possible
2) Another which behaves the way we expect in PC given the current state of the app

Love to get your thoughts on this @relgit

apizz:
Another issue: Tried to add `pfm_range_list` and `pfm_range_list_titles` for `VPNSubType`, given Apple Configurator 2 has several pre-configured known types.  Also added `pfm_range_list_allow_custom_value` to allow entering a custom string instead.  However, this appears to break things as ProfileCreator immediately crashes when the VPN payload is accessed.

Opted to include the known options as a pfm_note instead.

apizz:
Another issue: When a known VPN config bundle id is specified, `Role` (Vendor-Specific Role) is always added because a `pfm_require` value of `push` is specified.  This preference should only be included when either `net.juniper.sslvpn` or `net.pulsesecure.PulseSecure.vpnplugin` is specified, but this can't be manually removed.

When not applicable bundle id specified, which can't be removed:

Screen Shot 2020-08-23 at 9 44 35 PM

When applicable bundle id specified:

Screen Shot 2020-08-23 at 9 46 25 PM

This poses a challenge because not forcing this preference means for those who configure `net.juniper.sslvpn` or `net.pulsesecure.PulseSecure.vpnplugin` are missing something that's needed, while those who configure a different option are going to see a message for a preference that doesn't apply to them.  This isn't _terrible_, as ultimately the preference is excluded from the final profile but the logic (at least for me) is that if you see a warning it requires being dealt with.

Will likely need a pfm_note to address this.

kevinmcox:
@apizz @relgit 

Based on this comment:
https://github.com/ProfileCreator/ProfileCreator/issues/239#issuecomment-770566658
And this one:
https://github.com/ProfileManifests/ProfileManifests/issues/32#issuecomment-676368588

Can this Draft PR be closed?

apizz:
Yeah we can kill. It's in the history so can always give another go if we ever need

PR #362, opened 2020-09-16
Update com.microsoft.OneDrive.plist

walkintom:
Add "DisableTutorial" key, which is currently undocumented - https://macadmins.slack.com/archives/C1HNS6CTB/p1600202200065200


apizz:
Thanks.  Would love to get an app min version at some point.

PR #367, opened 2020-10-09
Add GUIDTool to RunScript

PeetMcK:
There's a function for both UID and GUID because some of use are oddly interested in such things ... i.e. I've had the same UID/GUID's for more than a decade and would like to keep them consistent. I asked and said feature was delivered.


apizz:
Thanks for this.  In future, be sure to at the very least update the `pfm_last_modified` date.  Otherwise, ProfileCreator won't know there's an update and therefore won't apply the change.  Not a big deal though, it's easy enough for me to do.

PeetMcK:
Super happy too! TBH I've only ever dug into the profile manifests enough to add the bits I need. Now that I know `pfm_last_modified `is there I'm happy to make sure it's updated! thanks.Peet

PR #369, opened 2020-10-24
Update Firefox for v81 & v82

apizz:
In Firefox 81, the format structure for preference subkeys within the larger 'Preferences' dictionary was changed such that each subkey is now its own dictionary.  This manifest has been updated to reflect this and the newly introduced 'Status' subkey was added for each.

Given the nature of this change, this would prevent admins from deploying the 'Preferences' dictionary to Macs for prior versions.

As part of these changes, the duplicate pfm_titles were also removed.

New preferences were included, but their value types have not been defined in either v2.3 or v2.4 of the Firefox policy releases.  This needs to be addressed before merging.

https://github.com/mozilla/policy-templates/issues/703


kevinmcox:
Closing due to age.

PR #403, opened 2021-01-06
Updated version check in index scripts to pass on Big Sur

relgit:
The index-update scripts stopped working on Big Sur because their version checks fail due to the versioning convention change on the OS (major version number no longer fixed at 10). This commit resolves the issue.

### DIFF
--- a/updateIndexIcons
+++ b/updateIndexIcons
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if (($( sw_vers -productVersion | awk -F. '{ print $2 }' ) < 14)); then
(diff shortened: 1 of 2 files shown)


apizz (changes requested):
I'm very confused by this change.  You print "10.14" and get a product version and then grep 10.14 so this always returns true?  To me this says nothing about what version the system is running and if it's greater than 10.14

relgit:
> You print "10.14" and get a product version and then grep 10.14 so this always returns true?

Almost. I print 10.14 along with the current version, and then sort them. If 10.14 comes out second after sorting this means that the current version is lower, therefore returning false.

apizz (approved):
Ah, I'm with you now.  That's clever.

PR #404, opened 2021-01-14
DigiDNA Jan 2021 tweaks

relgit:
The following commits all address keys that occurred twice in the same dictionary. Flat-out duplicates having the same value were removed, while those with different values were either combined under one key or replaced with the presumably intended PFM key.

### DIFF
--- a/Manifests/ManifestsApple/com.apple.applicationaccess-tvOS.plist
+++ b/Manifests/ManifestsApple/com.apple.applicationaccess-tvOS.plist
@@ -683,8 +683,6 @@ The payload organization for a payload need not match the payload organization i
 			<integer>1</integer>
 			<key>pfm_title</key>
 			<string>Deferred Software Updates Delay</string>
-			<key>pfm_tvos_min</key>
-			<string>12.2</string>
 			<key>pfm_type</key>
 			<string>integer</string>
 			<key>pfm_value_unit</key>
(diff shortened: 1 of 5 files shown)


apizz (commented):
I'm not sure I follow why you want to remove this key.  In the context of ProfileCreator, it displays the minimum os version below the key, which I personally find helpful and makes clear if a key is supported.

Manifests/ManifestsApple/com.apple.applicationaccess-tvOS.plist
```diff
@@ -683,8 +683,6 @@ The payload organization for a payload need not match the payload organization i
 			<integer>1</integer>
 			<key>pfm_title</key>
 			<string>Deferred Software Updates Delay</string>
-			<key>pfm_tvos_min</key>
-			<string>12.2</string>
```
apizz, line 687: Can we leave this one and remove the one on line 677?  Want to keep it in alphabetical order.

apizz (changes requested):
One minor change.  I like keeping things in alphabetical order so if we can do that in the other places as well just to keep it clean.

Manifests/ManifestsApple/com.apple.applicationaccess-tvOS.plist
relgit: I got ya.

apizz (approved):

PR #405, opened 2021-01-15
DigiDNA Proxy Manifests Updates

relgit:
The following commits bring the Global HTTP Proxy and Network Proxies up to date and aligned to the documentation.

PR #408, opened 2021-01-28
DigiDNA com.apple.print.add update

relgit:
This PR updates some of the texts of the `com.apple.print.add` manifest to better represent its role.

A note was also added to the Toolbar Items property, to warn admins about potential unwanted UI behavior that was reported by iMazing Profile Editor users and was successfully reproduced on our machines. This behavior persists after removal of the installing profile, although we did manage to restore to original state by either deleting the com.apple.print.add.plist file from the user preferences or by manually resetting the toolbar through the UI.

### DIFF
--- a/Manifests/ManagedPreferencesApple/com.apple.print.add.plist
+++ b/Manifests/ManagedPreferencesApple/com.apple.print.add.plist
@@ -141,7 +141,7 @@
 						<string>com.apple.print.add.search</string>
 					</array>
 					<key>pfm_description</key>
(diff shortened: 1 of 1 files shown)


apizz:
Just need to bump the `pfm_version` and update the last modified date @relgit and this looks good to merge

apizz (changes requested):
`pfm_version` & last modified date

relgit:
Good call; I've updated `pfm_last_modified` though kept `pfm_version` according to our discussion in #405.

PR #412, opened 2021-02-01
Add TouristDtoManifest script

apizz:
### Goal
- Take Apple's config.json and turn it into a manifest
- Have a good amount figured out (I think) but need a number of missing pieces:
  - Method for matching / associating the hardware model tour or other tour type to the correct manifest_segment_titles
  - Method for writing the pfm_name for all pref dictionaries to the manifest_segments dictionary
  - Likely more things because I haven't been able to get this working and output a .plist

[Recently accepted a PR ](https://github.com/ProfileCreator/macos-touristd-config/pull/1) to auto submit PRs when changes are detected.  Ideally would like to have this script likewise automatically run when a PR is merged from that repo and create our touristd manifest.

Haven't looked at this in a while, but if anyone with a little bit more python knowledge than me could help make this happen, 🥳

### DIFF
--- a/Resources/TouristdToManifest/TouristDToManifest.py
+++ b/Resources/TouristdToManifest/TouristDToManifest.py
@@ -104,10 +104,18 @@ def main():
 				model_type = model_type.replace( product_casing_key, product_casing[ product_casing_key ] )
 
 		# Parse title version
-		os_version = tour[ "osVersion" ]
-		os_version_fragments = os_version[ 0 ].split( "\\." )
-		os_version_fragments = list( filter( lambda fragment: ( fragment.isdigit() ), os_version_fragments ) )
-		title_os_version = ".".join( os_version_fragments )
+		if "osVersion" in tour:
+			os_version = tour[ "osVersion" ]
+			os_version_fragments = os_version[ 0 ].split( "\\." )
+			if len( os_version_fragments ) > 1:
+				# Version pattern is formatted like 10\\.15\\.[0-9]*
+				os_version_fragments = list( filter( lambda fragment: ( fragment.isdigit() ), os_version_fragments ) )
+				title_os_version = ".".join( os_version_fragments )
+			else:
+				# Version pattern is formatted like 10.15.* / 11.0*
+				os_version_fragments = os_version[ 0 ].rstrip( "*" ).split( "." )
+				os_version_fragments = list( filter( lambda fragment: ( fragment.isdigit() ), os_version_fragments ) )
+				title_os_version = ".".join( os_version_fragments )
 
 		new_subkey[ "pfm_title" ] = model_type + ": " + title_os_version
 
--- a/Resources/TouristdToManifest/TouristDToManifest.py
+++ b/Resources/TouristdToManifest/TouristDToManifest.py
@@ -121,7 +121,13 @@ def main():
 		else:
 			model_type = url_model_type.replace( "-", " " ).title()
 
-		new_subkey[ "pfm_title" ] = model_type
+		# Parse title version
+		os_version = tour[ "osVersion" ]
+		os_version_fragments = os_version[ 0 ].split( "\\." )
+		os_version_fragments = list( filter( lambda fragment: ( fragment.isdigit() ), os_version_fragments ) )
+		title_os_version = ".".join( os_version_fragments )
+
+		new_subkey[ "pfm_title" ] = model_type + ": " + title_os_version
 
 		# Add the new subkey to the new subkeys list
 		new_subkeys.append( new_subkey )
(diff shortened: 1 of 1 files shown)


relgit:
I've done some work on this adding the missing parts, and I need some input from you @apizz.

1. I'm not sure if the current manifest is ordered in any way, so the newly generated one simply follows the order of tours in the JSON file. If you know what order it should follow then we can change accordingly.
2. The `osVersion` key sometimes contains two versions. Do you know if there is logic on determining which one should be used?

One more point about my changes, is that I made it so that the old manifest is used as a template for the new one. This makes it easy to change meta information like payload description and the likes without going into the Python code and it makes the script cleaner; although I'm happy to revert if you'd like to keep the previous behavior.

apizz:
Thanks for working on this, @relgit !  I've approached the ordering in the following manner:

- Separate tabs / segments for each hardware and tour type, combining the New to Mac & What's New under a larger "macOS" heading.
- Within each tab / segment, order based on oldest to newest hardware and OS, but for models that have different size displays - like MacBook Pros and iMacs - put the smaller displays first.
  - For the MacBook Pros, I put the 13in. model before the 15 in. ordered by name, grouping OSes w/ oldest first, so this section would ideally start with the following `pfm_title`s:
    - MacBook Pro 13c: 10.12 - 10.14
    - MacBook Pro 13d: 10.12 - 10.14
    - MacBook Pro 15: 10.12 - 10.14
    - MacBook Pro 15d: 10.12 - 10.14
    - MacBook Pro 13: 10.13. - 10.14
    - MacBook Pro 15: 10.13 - 10.14
  - Just to illustrate the point further, for iMacs:
    - iMac 21a: 10.12 - 10.14
    - iMac 27: 10.12 - 10.14
    - iMac: 10.13 - 10.14
    - iMac 21a: 10.13. - 10.14
    - iMac: 10.14
    - iMac 21a: 10.14
    - iMac 21r: 10.14
    - iMac Early 2019: 10.14
  - For the macOS section, ideally would want the New to Mac preferences all grouped together, oldest to newest, followed by the What's New preferences grouped together, oldest to newest

apizz:
To your second point RE OS version, that's why I put OS range in the `pfm_title` .  So for the Mid 2018 MacBook Pro:

```
    {
      "boards": [
        "Mac-827FB448E656EC26",
        "Mac-937A206F2EE63C01"
      ],
      "iconAsset": "https://help.apple.com/macOS/configs/notifications/macbook-pro.icns",
      "waxwing": true,
      "id": "krdWS8DSQIqJSqQFXW1/pw",
      "platform": "mac",
      "minimumScale": 1,
      "type": 2,
      "osVersion": [
        "10\\.13\\.[1-9][0-9]*",
        "10\\.14\\.[0-9][0-9]*"
      ],
      "url": "https://help.apple.com/macOS/high-sierra/mid-2018/macbook-pro"
    },
```
That's how we get the `pfm_title` of `MacBook Pro Mid 2018: 10.13 - 10.14`

relgit (changes requested):
I was able to spend a few minutes improving the OS Version parsing so that it now supports ranges.

Now that we have the version ranges, we can tend to the sorting. The two secrets to accomplishing this are 2-column sorting and version range comparison. The latter is slightly more advanced, though if you feel confident you can try the former.

I'll try to get back to this soon if you're unsure.

apizz:
Thanks again for all your work on this @relgit.

Just ran this and it's very very close to being ready.  The range parsing is 💯 

One additional area I'd like to address is where more than one preference have the same pfm_title (ex. iMac: 10.15, MacBook Air: 10.15, etc.).  If we add extra processing such that the URL from the pfm_description is used to get `early-2020` or whatever the model release is we can append that to the pfm_title.

So at the end we'd do the following:
1. Recursively check for duplicate names within each section (iMac, iMac Pro, etc.) and see if we have preferences with duplicate pfm_titles.
2. If we find more than 1 with the same pfm_title, parse the pfm_description URL of each and do a regex search for 'early', 'mid', and 'late', since that's how Apple defines mid-year releases. ex. `https://help.apple.com/macos/catalina/early-2020/macbook-air`
3. Where we get a match on 'early', 'mid', or 'late', parse such that the matching URL component is added to the pfm_title: `MacBook Air: 10.15` is changed to `MacBook Air Early-2020: 10.15`

The last and final pieces relate to the macOS section - What's New & New to Mac:

1. New to Mac should be sorted with latest OS first (What's New is sorted this way)
2. Get To Know Your Mac Book appears to be a one-off in terms of URL formatting, which I think we can just change to `Get To Know Your MacBook`.

apizz:
Apple has shifted from the old touristd workflow to a new one with via "Tips" (com.apple.tipsd).  Therefore no reason for us to work on this.

Per others in the MacAdmin community, now that Tips is something which generates notifications, with a config profile that suppresses Tips notifications this effectively makes exploring and building out a more specific and robust tours manifest unneeded.

PR #452, opened 2021-07-20
Books - Added Welcome Screen Suppression for macOS 10.15+

walkintom:
Tested and working on macOS 10.15.7


walkintom:
Also proposing updating the name as the app is now just called Books rather than iBooks, although the payload domain remains unchanged.

PR #462, opened 2021-10-28
DigiDNA October 2021 tweaks

relgit:
This PR contains two housekeeping items:
* `pfm_required` keys changed to`pfm_require` (no trailing 'd') on preference keys where the associated value was appropriate for the latter
* Clarified lack of support for built-in filtering in the Web Content Filter manifest on macOS


apizz (approved):
Looks good, good catch(es)

PR #465, opened 2021-10-30
No need to make com.apple.dock "static-only" pref key an inverted value

bp88:
Verbiage is clear that if set to true the dock will be set to the static apps/docs and if set to false the static apps/docs will merge with the user's current dock icons. The inverted value currently causes confusion and unexpected behavior when you create the profile.


relgit:
Yes, this was my bad as I recently updated the wording on that key. @apizz with your blessing I will merge and update the index.

apizz:
@relgit if you would like to merge, go for it.  I've since moved on to Big Sur and ran into some issues trying to run the `updateManifestIndex` script. Did some tweaks but I'll try and get your 👀 on it in a separate PR.

relgit:
Thanks @bp88!

relgit:
@apizz there was a glitch in the manifest which I think caused `updateManifestIndex` to act up. All sorted now.

PR #472, opened 2021-11-29
Update updateIndexManifests to work on Big Sur

apizz:
- Confirmed works on M1 & Intel Macs running Big Sur
- Simplify index_path_tmp generation
- Add parens where appropriate
- Add timestamp var


homebysix:
Is there any interest in using GitHub Actions to generate/update the index files automatically? I could try to figure that out, if so.

apizz:
@homebysix that would be amazing!

homebysix:
Done: https://github.com/ProfileCreator/ProfileManifests/pull/482

PR #478, opened 2021-12-06
Add MunkiAdmin manifest

homebysix:
Resolves https://github.com/ProfileCreator/ProfileManifests/issues/85

This initial commit should cover all the settings available in the MunkiAdmin preferences UI, but there are still many hidden options not covered. I added a TODO section listing the keys to investigate in the future.


apizz:
Thanks so much!

PR #485, opened 2021-12-13
Remove default value for SlackDefaultSignInWorkspace

homebysix:
There is no default value for this setting, as it varies per organization.


kevinmcox:
Nice catch, thanks.

relgit:
@homebysix looks good; could you just also update the `pfm_last_modified` tag before we merge?

homebysix:
Done.

PR #489, opened 2021-12-18
Set all `pfm_format_version` to `<integer>1</integer>`

homebysix:
The [Manifest Format Versions](https://github.com/ProfileCreator/ProfileManifests/wiki/Manifest-Format-Versions) wiki page says:

> IMPORTANT
> These versions will be reset to 1 when the first non-beta release will be available. This is for testing the version system, and on release, the first version will include all keys available at that time.

This PR adjusts all instances of `pfm_format_version` to `<integer>1</integer>` to match that statement.

It's worth noting that the old [Apple preference manifest documentation](https://developer.apple.com/library/archive/documentation/MacOSXServer/Conceptual/Preference_Manifest_Files/Preface/Preface.html) defines this key (and other version keys) as `<real>1.0</real>`. I'm choosing to use integer to match conventions already in use in this repo.

I have not included updates to the `pfm_last_modified` or `pfm_version` but could add those if needed.

### DIFF
--- a/Manifests/ManagedPreferencesApple/com.apple.Safari.SandboxBroker.plist
+++ b/Manifests/ManagedPreferencesApple/com.apple.Safari.SandboxBroker.plist
@@ -11,7 +11,7 @@
 	<key>pfm_domain</key>
 	<string>com.apple.Safari.SandboxBroker</string>
 	<key>pfm_format_version</key>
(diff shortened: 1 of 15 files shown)


homebysix:
Added last modification date updates.

apizz (approved):

PR #497, opened 2021-12-25
Add additional Zoom preferences

apizz:


kevinmcox:
It looks like there are some inconsistencies in the Zoom documentation:
https://support.zoom.us/hc/en-us/articles/115001799006-Mass-deploying-preconfigured-settings-for-Mac

For instance in the example XML near the top of the page they show "nogoogle" and "nofacebook" as the keys. However they are listed as "NoGoogle" and "NoFacebook" in the table further down the page.

I wonder if one of those is wrong or if the capitalization doesn't actually matter for some of those keys?

apizz:
@kevinmcox recently a colleague of mine was testing Zoom configs using the documented capitalization rather than the previous lowercase and it works as expected.  Maybe you're right and capitalization doesn't matter, but going to leave as documented for consistency.

I've already reported 1 issue on the page to Zoom support: `SetScreenCaptureMode` is in the 'Meetings' section which indicates at the top of the section that all values are boolean, but that's just not possible given there are 5 possible values [0-4].  In my testing confirmed that this is in fact a `string` and not an `integer`.

PR #503, opened 2022-01-12
DigiDNA Parental Controls: Application Access improvements

relgit:
Added missing column titles and made conditional exclusions support value-less targets. Also, added the documentation URL.


kevinmcox (commented):
Looks good to me.

Issue #377, opened 2020-11-10
com.apple.applicationaccess-macOS.plist


bp88:
### Please provide the following information:

**Name of App / Payload**: *com.apple.applicationaccess*

**Preference key(s) needing fixing**: 

**Description of problem(s) / error(s) with listed key(s)**:
*Open up the Restrictions payload for macOS, the following tabs are completely empty when you click on them:
AirPrint, Classroom and Updates*

**Description of expected behavior**:
I expect to see some preferences in these tabs.

**Applicable Documentation URL(s)**:

**How are you testing / deploying your profile?**: Running ProfileCreator 0.3.2

apizz:
After submitting PR #387 seeing updates missing as well.  Pretty sure this is a subdomain thing.

![Screen Shot 2020-11-21 at 5 36 23 PM](https://user-images.githubusercontent.com/11789931/99889083-1baf9700-2c20-11eb-99f5-182858c11307.png)

apizz:
@erikberglund if you're able, can you take a look at this?

kevinmcox:
There is still an issue here. The tabs are no longer empty, but AirPrint and Classroom display the same single key, `enforcedFingerprintTimeout` in Profile Creator, but completely different options in iMazing.

Is there a limitation in PC that causes this while iMazing can handle it? @relgit can you take a look since you've been working in the other Restrictions profiles recently?

relgit:
Sure. Looks like `enforcedFingerprintTimeout` was not assigned to a segment. I've just added deae6f7 in #504 which fixes that.

PR #504, opened 2022-01-12
DigiDNA Restrictions tweaks

relgit:
This PR makes two miscellaneous tweaks in the Restrictions manifest:
* Removes redundant appearance of key name in segment
*  Adds default value to key


kevinmcox (commented):
Dang this is a massive profile...

Despite having an `AirPlay` section in `pfm_segments` of both the old and updated manifest, I do not get an AirPlay section in the GUI.

relgit:
It's a beast, I know.

Looks like `pfm_range_list_titles` omits AirPlay for some reason. Let me add a commit amending this quickly.

relgit:
There; try when you're ready.

relgit:
Apart for corrections made after discussion in #377, I've now also added a commit removing key-level documentation URLs in two cases where they were equal to the domain's general one.

apizz (approved):

PR #507, opened 2022-01-23

kevinmcox:
• Adds new `UpdaterOptimization` key
• Changes Teams ID to new value
• Sets `ChannelName` to be required if `ManifestServer` is configured


apizz:
Thanks for this @kevinmcox .  FYI - if the `pfm_range_list_titles` has no variance from `pfm_range_list` then it's not necessary.

Also, was looking at https://downloads.lib.utah.edu/media_streaming_presentation_documents/pdf/mac_mgrs/20220119_mm/2022.01.19_mm_office.pdf and it appears there's also support for updating TEAMS10 with MAU 4.45 and wanted to make sure that was included here.

kevinmcox:
That makes since on the range list, thanks.

I changed the Teams Application ID already for that upcoming support.

PR #509, opened 2022-01-27
DigiDNA self-exclusion removed in the Passcode manifest

relgit:
This PR removes all conditionals that caused keys to exclude themselves in the Passcode manifest.

I believe that these conditions were some sort of an early attempt at preventing write-out of default values, at least in the case of `minLength` and `minComplexChars`. Since we do not prevent admins from reaffirming defaults, the conditions should be removed.

As per `maxInactivity` which has no default value, I can venture a guess that the condition was probably left there by mistake after copy-and-pasting, since it incorrectly prevented a value of `0` from being set on the key.

As a bonus, I've updated the conditional on `maxFailedAttempts` to explicitly represent requirement.


apizz (approved):
lgtm

PR #516, opened 2022-02-02
 Update macOS restrictions correction

relgit:
Just adding a small correction to the title of a key added in #515.


apizz (approved):
Thanks for correcting my mistak

PR #524, opened 2022-02-14
DigiDNA MCX Wi-Fi Managed correction

relgit:
The Wi-Fi Managed Settings manifest uses an incorrect preference domain when it should just be a subdomain of the greater MCX domain.

This PR renames the manifest file according to convention used on other subdomains of MCX and corrects the information inside the manifest to match. It also contains two minor rewording tweaks.

### DIFF
--- a/Manifests/ManifestsApple/com.apple.MCX-WiFiManagedSettings.plist
+++ b/Manifests/ManifestsApple/com.apple.MCX-WiFiManagedSettings.plist
@@ -3,28 +3,28 @@
 <plist version="1.0">
 <dict>
 	<key>pfm_description</key>
-	<string>Wi-Fi Managed Settings</string>
+	<string>Use this section to define managed Wi-Fi settings</string>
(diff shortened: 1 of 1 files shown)


kevinmcox (commented):
Looks OK to me.

apizz (approved):

apizz (changes requested):
Sorry we need a pfm_last_modified update pls & ty

relgit:
It's buried, but it's there.

kevinmcox (approved):

Reem502 (approved):

apizz (approved):

apizz:
I don't know who @Reem502 is but given they're a new user and only those who have some role in this project should be able approve or request changes on code review, I've enabled a setting to lock this down.

relgit:
I don't know either. Perhaps someone who had just joined GitHub as you said and who doesn't yet know their way around the site 🤷‍♂️

PR #532, opened 2022-02-23
Added Install or Defer manifest

mpanighetti:
- added manifest for Install or Defer (https://github.com/mpanighetti/install-or-defer#configuration-profile)

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.github.mpanighetti.install-or-defer.plist
+++ b/Manifests/ManagedPreferencesApplications/com.github.mpanighetti.install-or-defer.plist
@@ -179,7 +179,7 @@ A profile can consist of payloads with different version numbers. For example, c
 			<key>pfm_name</key>
 			<string>DiagnosticLog</string>
 			<key>pfm_title</key>
-			<string>Diagnostic Log</string>
+			<string>Enable Diagnostic Log</string>
 			<key>pfm_type</key>
 			<string>boolean</string>
 		</dict>
--- a/Manifests/ManagedPreferencesApplications/com.github.mpanighetti.install-or-defer.plist
+++ b/Manifests/ManagedPreferencesApplications/com.github.mpanighetti.install-or-defer.plist
@@ -124,6 +124,8 @@ A profile can consist of payloads with different version numbers. For example, c
 		<dict>
 			<key>pfm_app_min</key>
 			<string>5.0</string>
+			<key>pfm_default</key>
+			<string>Install</string>
 			<key>pfm_description</key>
 			<string>The label of the install button.</string>
 			<key>pfm_description_reference</key>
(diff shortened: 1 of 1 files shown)


kevinmcox:
@mpanighetti Thanks for the contribution!

Looking at the Release History it looks like you could also use app minimim keys on a number of the preferences. This highlights that they must be on a specific release of Install or Defer for the keys to work (at Line 125 for example) e.g.
```
<key>pfm_app_min</key>
<string>5.0</string>
```

It can get a little repetitive, but if you add this to the individual payload keys it'll make the link to the documentation show up in the ProfileCreator GUI for easy reference (at Line 129 for example):

```
<string>https://github.com/mpanighetti/install-or-defer#configuration-profile</string>
```

You probably also want to specify the targets: system, user or both? (Line 343)
```
<key>pfm_targets</key>
<array>
	<string>system</string>
	<string>user</string>
</array>
```

kevinmcox (changes requested):
@mpanighetti Thanks for the contribution!

Looking at the Release History it looks like you could also use app minimim keys on a number of the preferences. This highlights that they must be on a specific release of Install or Defer for the keys to work (at Line 125 for example) e.g.
```
<key>pfm_app_min</key>
<string>5.0</string>
```

It can get a little repetitive, but if you add this to the individual payload keys it'll make the link to the documentation show up in the ProfileCreator GUI for easy reference (at Line 129 for example):

```
<string>https://github.com/mpanighetti/install-or-defer#configuration-profile</string>
```

You probably also want to specify the targets: system, user or both? (Line 343)
```
<key>pfm_targets</key>
<array>
	<string>system</string>
	<string>user</string>
</array>
```

mpanighetti:
Thanks! Made those adjustments (I'll update the docs on the app side to specify minimum version for each key for consistency), let me know if anything else should be changed.

kevinmcox (approved):

apizz:
@mpanighetti I made some tweaks.  At this time, the `pfm_description_reference` key isn't visible in the UI, so took some info that comes across as helpful or important missing from the `pfm_description` and put in a `pfm_note`.  Check out my notes in the 'Tweaks' commit.

If the changes look OK to you, will get this merged.

Thanks again for the contribution!

apizz (approved):

mpanighetti:
Changes look good to me, thanks for your help!

Manifests/ManagedPreferencesApplications/com.github.mpanighetti.install-or-defer.plist
```diff
@@ -0,0 +1,410 @@
...
+			<key>pfm_description</key>
+			<string>Whether to write to a persistent log file at /var/log/install-or-defer.log.</string>
+			<key>pfm_description_reference</key>
+			<string>Whether to write to a persistent log file at /var/log/install-or-defer.log. Defaults to false, instead writing all output to the system log for live diagnostics.</string>
+			<key>pfm_documentation_url</key>
+			<string>https://github.com/mpanighetti/install-or-defer#configuration-profile</string>
+			<key>pfm_name</key>
+			<string>DiagnosticLog</string>
+			<key>pfm_title</key>
+			<string>Diagnostic Log</string>
```
relgit, line 182: Can this be reworded into _Persist Diagnostic Log_? There's no description for booleans in IPE so this would give a bit more context.

relgit (changes requested):
Other than my one comment this looks good to me.

Manifests/ManagedPreferencesApplications/com.github.mpanighetti.install-or-defer.plist
mpanighetti: Renamed to "Enable Diagnostic Log" to better indicate it's a boolean key.

PR #537, opened 2022-02-24
Add allowApplePersonalizedAdvertising to macOS Restrictions

apizz:


apizz:
I'M SILLY - wrong payload: com.apple.AdLib, we already added this...

PR #542, opened 2022-03-02
Change Fast User Switching default

kevinmcox:
Discussion in Slack made us realize that the default behavior of `MultipleSessionEnabled` is actually `false` contrary to Apple's documentation.


kevinmcox:
Additional testing and discussion in Slack revealed that the default behavior for this key varies widely depending on the number of users on a computer and how they were created (System Preferences vs. MDM).

Closing this in lieu of a different solution.

PR #550, opened 2022-06-08
Add Network Share Mounter and icon

doschupp:
Network Share Mounter is a macOS utility to automatically mount network shares. It has arisen from the need of our users at the university to automatically mount department/usergroup specific SMB shares. Shares are configurable via Profiles (i.g. MDM) or by a user himself. 

In the last few weeks we have received many questions about how to configure / create a profile for the app. So a ProfileCreator manifest should simplify this! :-)

Project: https://gitlab.rrze.fau.de/faumac/networkShareMounter


doschupp:
Thanks @relgit for your help. I have implemented your recommendations.

Yes, the `customNetworkShares` value isn't implemented because this array will be used for the custom shares added by the users in settings (right image). It is possible to set the value via a profile, but not recommend.

PR #551, opened 2022-06-24
added new install-or-defer keys

mpanighetti:
- added new install-or-defer keys from versions 5.0.3, 5.0.4, and 5.0.5 (see https://github.com/mpanighetti/install-or-defer#configuration-profile for current docs)
- reorganized keys to match docs (now grouped by category)

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.github.mpanighetti.install-or-defer.plist
+++ b/Manifests/ManagedPreferencesApplications/com.github.mpanighetti.install-or-defer.plist
@@ -13,7 +13,7 @@
 	<key>pfm_format_version</key>
 	<integer>1</integer>
 	<key>pfm_last_modified</key>
-	<date>2022-02-23T07:10:46Z</date>
+	<date>2022-06-24T22:27:00Z</date>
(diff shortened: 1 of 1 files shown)


relgit (changes requested):
Thanks for the updates, @mpanighetti.

I'm happy to merge this, but before I do would you mind removing the `pfm_documentation_url` tags from the subkeys? There's no need to have them if their URL is the same as on the root.

Cheers!

relgit (approved):
Indeed, we're giving you contradicting guidance; sorry about that! I'll make sure we agree on one preferred style.

For now though, thanks for making the changes.

apizz:
@relgit the reason we add the documentation URL to each key is because when this is configured there's a little book icon in ProfileCreator which will open the referenced URL.  If this is only set at the root-level this won't be accessible in the app.

If we can leave the key-level `pfm_documentaton_url` items in place that would be helpful

mpanighetti:
Sounds good, I've restored the per-entry key.

relgit:
If there's no root-level display currently in PC then, yes, there's no other choice.

PR #554, opened 2022-07-29
Update com.apple.loginwindow.plist

walkintom:
Added "TALLogoutSavesState" key

### DIFF
--- a/Manifests/ManifestsApple/com.apple.loginwindow.plist
+++ b/Manifests/ManifestsApple/com.apple.loginwindow.plist
@@ -14,7 +14,7 @@ window payloads may be installed together.</string>
 	<key>pfm_interaction</key>
 	<string>combined</string>
 	<key>pfm_last_modified</key>
-	<date>2022-07-29T12:26:09Z</date>
+	<date>2022-07-29T16:07:04Z</date>
 	<key>pfm_platforms</key>
 	<array>
 		<string>macOS</string>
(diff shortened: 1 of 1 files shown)


Manifests/ManifestsApple/com.apple.loginwindow.plist
```diff
@@ -608,6 +609,22 @@ Availability: Available in macOS 10.13 and later.</string>
 			<key>pfm_type</key>
 			<string>boolean</string>
 		</dict>
+		<dict>
+			<key>pfm_default</key>
+			<true/>
+			<key>pfm_description</key>
+			<string>Reopens windows that were open at time of logout</string>
+			<key>pfm_description_reference</key>
```
kevinmcox, line 617: No reason to leave in keys with empty strings.

Manifests/ManifestsApple/com.apple.loginwindow.plist
```diff
@@ -608,6 +609,22 @@ Availability: Available in macOS 10.13 and later.</string>
...
+			<string></string>
+			<key>pfm_documentation_source</key>
```
kevinmcox, line 619: No reason to leave in keys with empty strings.

kevinmcox (changes requested):
Does this key work alone? In the past when I configured this option via script I was also changing the `LoginwindowLaunchesRelaunchApps` key (but reports vary on if it is necessary).

Manifests/ManifestsApple/com.apple.loginwindow.plist
walkintom: Must admit, I never even knew the `LoginwindowLaunchesRelaunchApps` key existed so yeah, `TALLogoutSavesState` does work independently. I'll fix the empty keys.

relgit (approved):

kevinmcox (approved):

PR #558, opened 2022-08-12
Added Initial Manifest for XCreds

twocanoes:
Profile Manifest for macOS app XCreds for using your cloud credentials to login and sync your local password with your mac.


apizz:
Thanks for this @twocanoes .  I took the liberty of clearing a bunch of unnecessary whitespace and comments in the manifest, but there a few things I'd to see get addressed before I merge:

1. Please sort the keys in each preference dictionary alphabetically.  This makes it easier to read & manage when future changes are needed.
2. Please add details to `pfm_description` for each preference.  This should provide basic information about what the preference does.  You've essentially done that her e(https://github.com/twocanoes/xcreds/wiki/AdminGuide#preferences) but this information should be included in the manifest so that other admins know exactly what something does & what this should be set as.  You can also use `pfm_note` if you need another area to provide additional info / details
3. Remove any blank keys in each preference dictionary.  for example, there are a number of `pfm_default` keys which have no value.  If there is no default value or other keys that provide no value, please remove these.

While not critical, it would also be helpful if on your https://github.com/twocanoes/xcreds/wiki/AdminGuide#preferences page if you turned each preference name into a header.  This way you could more easily link to these and thereby also add a `pfm_documentation_url` for each preference to provide admins a quick way to get that information for more context.

Let me know if you have any questions.

Thanks for contributing!

apizz (changes requested):
See above comments

twocanoes:
added changes requested. Let me know if I messed it up :).

apizz:
1. https://github.com/twocanoes/xcreds/wiki/AdminGuide#enablefderekey indicates this is a `string`, but appears to be configured as a `boolean`, which makes sense.  Think the error is on the GitHub wiki reference.
2. https://github.com/twocanoes/xcreds/wiki/AdminGuide#logfilename references `~/Library/Logs/` but the manifest indicates `/tmp`.  Which is correct?
3. RE the above pref, can a different path be specified, or it's just the name of the file as the preference name suggests?
4. https://github.com/twocanoes/xcreds/wiki/AdminGuide#username indicates it's a `boolean` but the manifest says it's a `string`, which makes sense. Think the error is on the GitHub wiki reference.
5. https://github.com/twocanoes/xcreds/wiki/AdminGuide#enablefderecoverykeypath indicates it's a `boolean` but the manifest says it's a `string`, which makes sense. Think the error is on the GitHub wiki reference.

apizz:
Also, is there any reason you chose to make `refreshRateHours` a string instead of an integer?  Reason being, there's a nifty option - `pfm_range_min` - which if the preference type was `integer` we could specify a value of `1` to lock in that minimum value so anything less than this is flat out rejected.  Related, is there a max value?  

What this would look like in the manifest:
```
<dict>
	<key>pfm_default</key>
	<integer>3</integer>
	<key>pfm_description</key>
	<string>Number of hours for checking for password changes. Default is 3 hours.  Minimum is 1 hour.</string>
	<string>https://github.com/twocanoes/xcreds/wiki/AdminGuide#refreshratehours</string>
	<key>pfm_name</key>
	<string>refreshRateHours</string>
	<key>pfm_range_min</key>
	<integer>1</integer>
	<key>pfm_title</key>
	<string>Password Change Check Rate</string>
	<key>pfm_type</key>
	<string>integer</string>
	<key>pfm_value_unit</key>
	<string>hours</string>
</dict>
```

twocanoes:
I updated the wiki to address #1, #4, and #5.

I removed logfilename pref to address #2 and #3. It really shouldn't be set.

I changed refreshRateHours to be an int and have a max number.

Thanks for all the cleanup!

apizz (approved):

PR #565, opened 2022-09-29

doschupp:
Change regex value for network shares to allow spaces in the pathname after the server address


relgit (changes requested):
Please update the manifest's `pfm_last_modified` tag, thanks!

relgit (approved):

apizz (approved):

PR #568, opened 2022-10-10
DigiDNA October 2022 updates

DigiYann:
Multiple fixes and updates to various manifests


Manifests/ManifestsApple/com.apple.dnsProxy.managed.plist
```diff
@@ -183,8 +182,6 @@ The payload organization for a payload need not match the payload organization i
 			<string>plist2dict</string>
 		</dict>
 	</array>
-	<key>pfm_supervised</key>
```
kevinmcox, line 186: This line from the documentation doesn't make sense to me:

"Beginning with iOS 15, this profile is unsupervised and must be installed via MDM."

relgit: Yeah it should have said _Beginning with iOS 15, this payload no longer requires a supervised device but must be installed via MDM_.

apizz (approved):
lgtm

PR #571, opened 2022-10-31
Brave-specific Preference Fix

apizz:
Originally communicated in Jamf manifest mirror by @dustinjlandgraf: https://github.com/Jamf-Custom-Profile-Schemas/ProfileManifestsMirror/pull/14

Admittedly, could only confirm 3 of the 4 listed prefs on https://support.brave.com/hc/en-us/articles/360039248271-Group-Policy are boolean as `BraveWalletDisabled` wouldn't display in brave://policy, even after quitting Brave and running `sudo killall cfprefsd`.

Screen Shot 2022-10-31 at 7 19 12 PM

It is annoying that we keep getting these preferences which have 'disabled' or 'enabled' in the key name, rather than a standard use of one of these so that we can always know that `false` = disabled or that `true` = enabled. ¯\_ (ツ)_/¯

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.brave.Browser.plist
+++ b/Manifests/ManagedPreferencesApplications/com.brave.Browser.plist
@@ -15,7 +15,7 @@
 	<key>pfm_interaction</key>
 	<string>combined</string>
 	<key>pfm_last_modified</key>
-	<date>2021-12-23T00:31:31Z</date>
+	<date>2022-10-31T23:14:14Z</date>
 	<key>pfm_platforms</key>
 	<array>
 		<string>macOS</string>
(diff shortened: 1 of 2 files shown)


relgit:
@apizz how committed are you to solving this with `pfm_range`? I share your sentiment, but personally I prefer keeping these as check-boxes, instead of popup-buttons because those will add an additional click.

I think that the choices that you made for title texts are perfect and should clear up any confusion.

P.S. I'm glad to see the ChromeToBrave script being used as we envisioned with the domain specific Plist.

dustinjlandgraf:
My apologies for posting the change to the mirror. I realized it after submitted. I'm glad it found it's way here though.

apizz:
@relgit I have no personal preference. For transparency, I only copied & pasted what was already in the existing .plist.  Can make these checkboxes instead.

apizz:
No apologies necessary, @dustinjlandgraf .  @relgit let me know what you think after these changes.

And yes, the script & .plist template is _super_ helpful :) Makes this stuff so much more manageable

relgit (approved):
Works for me, thanks.

PR #578, opened 2022-11-12
Allow the use of Value Importers for system extension policy

WillYu91:
PR for supporting the changes in ProfileCreator here https://github.com/ProfileCreator/ProfileCreator/pull/254

### DIFF
--- a/Manifests/ManifestsApple/com.apple.system-extension-policy.plist
+++ b/Manifests/ManifestsApple/com.apple.system-extension-policy.plist
@@ -204,6 +204,8 @@
 			<string>Allowed System Extensions</string>
 			<key>pfm_type</key>
 			<string>dictionary</string>
+			<key>pfm_value_import_processor</key>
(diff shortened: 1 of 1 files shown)


Manifests/ManifestsApple/com.apple.system-extension-policy.plist
```diff
@@ -204,6 +204,8 @@
 			<string>Allowed System Extensions</string>
 			<key>pfm_type</key>
 			<string>dictionary</string>
+			<key>pfm_value_import_processor</key>
```
apizz, line 207: Whenever you update a manifest, the pfm_last_modified date at the top of the manifest needs to be updated.  Only when preferences are added / removed do you also then need to bump the pfm_version key.

apizz (changes requested):

Manifests/ManifestsApple/com.apple.system-extension-policy.plist
WillYu91: Done

PR #579, opened 2022-11-13
Update 1Password 8 Manifest

fgeronimo-panther:
Per @apizz [comment](https://github.com/ProfileCreator/ProfileManifests/pull/577#discussion_r1019827300), I forgot to commit this change in this [PR](https://github.com/ProfileCreator/ProfileManifests/pull/577) that the 1Password 8 title and description in the preference domain has been standardized to not be version specific, we remove the version number.

- Remove 1Password version number from manifest
- Tested on ProfileCreator 0.3.3

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.1password.1password.plist
+++ b/Manifests/ManagedPreferencesApplications/com.1password.1password.plist
@@ -15,7 +15,7 @@
 	<key>pfm_format_version</key>
 	<integer>1</integer>
 	<key>pfm_last_modified</key>
-	<date>2022-11-10T23:29:11Z</date>
+	<date>2022-11-13T23:26:23Z</date>
 	<key>pfm_platforms</key>
 	<array>
 		<string>macOS</string>
--- a/Manifests/ManagedPreferencesApplications/com.1password.1password.plist
+++ b/Manifests/ManagedPreferencesApplications/com.1password.1password.plist
@@ -7,7 +7,7 @@
 	<key>pfm_app_url</key>
 	<string>https://1password.com</string>
 	<key>pfm_description</key>
-	<string>1Password 8 settings</string>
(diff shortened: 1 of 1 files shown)


Manifests/ManagedPreferencesApplications/com.1password.1password.plist
```diff
@@ -7,7 +7,7 @@
 	<key>pfm_app_url</key>
 	<string>https://1password.com</string>
 	<key>pfm_description</key>
-	<string>1Password 8 settings</string>
+	<string>1Password settings</string>
```
apizz, line 10: Also need to bump the pfm_last_modified date, but thanks for fixing this.

apizz (changes requested):

Manifests/ManagedPreferencesApplications/com.1password.1password.plist
fgeronimo-panther: Fixed in [commit](https://github.com/ProfileCreator/ProfileManifests/pull/579/commits/b772ca633cf4bed43aa0f10a79ab457b21cd9ad4)

Manifests/ManagedPreferencesApplications/com.1password.1password.plist
```diff
@@ -7,15 +7,15 @@
 	<key>pfm_app_url</key>
 	<string>https://1password.com</string>
 	<key>pfm_description</key>
```
apizz, line 9: Let's also add `pfm_app_min` to the root-level manifest so we're documenting formally that this manifest requires v8.

Given when MDM support was added we probably should add `pfm_add_min` to the individual preferences as well so if and when new management features are added we can build off of that.

apizz (changes requested):

apizz (approved):

PR #581, opened 2022-11-28
Add VPN Tracker Enterprise VPN Profile Manifest

adrianthomas:
We've added support for Profiles in VPN Tracker on Mac and iOS and wanted to share the manifest with the #MacAdmins community. 
Thanks!


relgit (changes requested):
Thanks for submitting this work, @adrianthomas.

Could you run this manifest through the command `plutil -convert xml1 com.vpntracker.365mac-config.plist`? This will remove spaces between key-value pairs, and will sort the keys alphabetically.

Then, you could if you wish, add a `pfm_exclusion` structure so that `Managed_Onboarding_FullName` gets excluded when `Managed_Onboarding_GivenName` and `Managed_Onboarding_FamilyName` are present. That's up to you.

Manifests/ManagedPreferencesApplications/com.vpntracker.365mac-config.plist
```diff
@@ -0,0 +1,344 @@
...
+	0ilHMPUgIf4LyUdDjYLrcT2uR3Xx71fhewWLJrR8AAAAAElFTkSuQmCC
+	</data>
+	<key>pfm_last_modified</key>
+	<date>2022-11-11T09:00:00Z</date>
+	<key>pfm_platforms</key>
+	<array>
+		<string>macOS</string>
+	</array>
+	<key>pfm_subdomain</key>
+	<string>Onboarding</string>
```
apizz, line 124: Is this intentional?  https://github.com/ProfileCreator/ProfileManifests/wiki/Subdomains

Manifests/ManagedPreferencesApplications/com.vpntracker.365mac-config.plist
```diff
@@ -0,0 +1,344 @@
...
+	<key>pfm_app_url</key>
+	<string>https://www.vpntracker.com</string>
+	<key>pfm_description</key>
+	<string>Manage settings in VPN Tracker for Mac, iPhone or iPad – the multi protocol VPN client for enterprise teams. Visit vpntracker.com/goto/profile to learn more.</string>
+	<key>pfm_domain</key>
+	<string>com.vpntracker.365mac</string>
+	<key>pfm_format_version</key>
+	<integer>6</integer>
+	<key>pfm_icon</key>
+	<data>
```
apizz, line 14: This _shouldn't_ be necessary.  Will auto load an icon of the same name as the manifest in `Icons/ManagedPreferenfcesApplications/<name_of_manifest.png`.  Should be 64x64

Manifests/ManagedPreferencesApplications/com.vpntracker.365mac-config.plist
```diff
@@ -0,0 +1,344 @@
...
+			<key>pfm_type</key>
+			<string>string</string>
+			<key>pfm_value_placeholder</key>
+			<string>Grace Hopper</string>
+		</dict>
+	</array>
+	<key>pfm_targets</key>
+	<array>
+		<string>user</string>
+	</array>
```
apizz, line 336: Are you certain that this is only in the user-level profile context?

apizz (changes requested):

Manifests/ManagedPreferencesApplications/com.vpntracker.365mac-config.plist
DarkDust: It was intentional as I was planning ahead: we might want to add additional MDM support and I thought it would be a good idea to have separate manifests for onboarding (user management) and other, unrelated settings. Should we remove it?

Manifests/ManagedPreferencesApplications/com.vpntracker.365mac-config.plist
DarkDust: No, I'm not. I tried reading up what this value actually means but failed to find a good explanation. As far as I understand, `system` is also called device-level and means that a profile can be installed system-wide and applies to all users, while `user` just targets a specific user and does not affect other users on a system. Is that correct? If so, both levels would probably be applicable to our app.

Manifests/ManagedPreferencesApplications/com.vpntracker.365mac-config.plist
DarkDust: Thank you, will provide an icon file with our next push.

Manifests/ManagedPreferencesApplications/com.vpntracker.365mac-config.plist
apizz: It's more about user experience.  The reason we had to implement this feature is that there are a number of profile payloads like Energy Saver (https://developer.apple.com/documentation/devicemanagement/energysaver) and Time (https://developer.apple.com/documentation/devicemanagement/timeserver) - 2 completely different and unrelated things - but which Apple ultimately has handled in the same payload: `com.apple.MCX`.

All that the `pfm_subdomain` key does this allows us to do is treat these (and related) items as _separate_ manifests & payloads in the interface but which ultimately export to the same `com.apple.MCX`.

So, if you want admins to do that in this case that's one thing, but given how you've named the plist - `com.vpntracker.365mac-config.plist` - and that you've just specified 'Onboarding', I'm not sure what your intent is there.  Ultimately, if this is really all going to center around the same core thing (VPN Tracker) then you probably just want to have it all within 1 manifest and separate out the settings into separate menus (Zoom and a number of other payloads do this to group settings together to make things easier to find and not just be a long list of options).

Manifests/ManagedPreferencesApplications/com.vpntracker.365mac-config.plist
apizz: That is correct.  FWIW, I'm not entirely sure in this day and age why someone would configure a profile to be user-specific, because at the end of the day if it's your org's device you probably want anyone within your org to use it.

So you can go ahead and add the `<string>system</string>` option here as well.

DarkDust:
I'm sorry it took so long, I've updated the manifest and extracted the icon.

DarkDust:
Hi @relgit , @apizz , is there anything else you need from us for this PR?

relgit (approved):

PR #585, opened 2022-12-04
Fix Printing Manifest UserPrinterList

### LINKED ISSUE #584: com.apple.mcxprinting does not allow multiple printers...

**Name of App / Payload**: com.apple.mcxprinting 
**Preference key(s) needing fixing**: UserPrinterList

**Description of problem(s) / error(s) with listed key(s)**: *ex. When `X` preference is configured, it does not *
Does not allow for multiple printers in UserPrinterList per 
https://developer.apple.com/documentation/devicemanagement/printing

**Description of expected behavior**:
Multiple printers should be able to be configured.

**Applicable Documentation URL(s)**: 
https://developer.apple.com/documentation/devicemanagement/printing

**How are you testing / deploying your profile?**: *ex. manual profile install, MDM, etc.*
Both manually and via Jamf Pro MDM  with XML and JSON versions.

apizz:
Resolves #584 

Was not correctly setup to be an array of dictionaries.  Changes correctly show this in the UI.

Screenshot 2022-12-03 at 11 25 00 PM

Also:
- Add a `pfm_exclude` for `DefaultPrinter` since this needs to be in the UserPrinterList
- Update some descriptions for clarity

### DIFF
--- a/Manifests/ManifestsApple/com.apple.mcxprinting.plist
+++ b/Manifests/ManifestsApple/com.apple.mcxprinting.plist
@@ -15,7 +15,7 @@
 		<key>pfm_interaction</key>
 		<string>exlusive</string>
 		<key>pfm_last_modified</key>
-		<date>2022-10-10T08:08:05Z</date>
+		<date>2022-12-04T04:14:48Z</date>
 		<key>pfm_macos_min</key>
 		<string>10.7</string>
 		<key>pfm_platforms</key>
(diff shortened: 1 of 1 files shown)


kevinmcox (approved):
Looks good, thanks for the fix @apizz.

relgit:
@apizz, @kevinmcox, I think this needs another look. The key is documented as a dictionary and changing it to array is incorrect (@DigiYann ran tests yesterday to confirm).

The previous manifest allowed entering one printer, and in the profile this was represented by a dictionary where the configured printer appeared under the key `Printer`. In reality this key was more akin to an identifier than to a preference key because it could be anything, and `Printer` was just an example provided by the manifest. Adding more printers was possible in a text editor provided that each also had its own key.

The conventional way to represent this in the manifest is by changing the example key name from `Printer` to `{{key}}`, but that would do us no good at all because such a structure is so rare that neither iMazing Profile Editor nor ProfileCreator support it right now.

I've considered suggesting that we relax the `pfm_unique` restriction so that users could add multiple printers by adding multiple payloads but testing showed this approach to fail as well.

Perhaps you guys have other ideas to try, but otherwise I recommend that we revert this PR and keep the issue open.

apizz:
The convention of defining an array of dictionaries has been in use within ProfileCreator manifests for some time: https://github.com/ProfileCreator/ProfileManifests/wiki/Array-of-Dictionaries

We can't simply assume that a dictionary preference which has a pfm_subkey with a preference that is a type of dictionary that this is an array of dictionaries.

In this case, the user printer list preference key is irrelevant because it's simply a container for the actual preferences an administrator would want to configure: the printers they want to have deployed via the profile. So I would argue given this has been used in other preferences involving arrays of dictionaries that you either incorporate this into iMazing profile editor or we determine a different way of configuring array of dictionaries in manifests.

Given the user exp. now matches what it should have been, we won't be reverting this change.

everetteallen:
@apizz is anyone working on this?

PR #586, opened 2022-12-14
Add Missing Zoom Preferences & Auto Update v2

### LINKED ISSUE #583: Add Zoom AU2 autoupdate and additional prefs
**Name of App / Payload**: Zoom

**Preference key(s) to be added / updated**: AU2_* prefs

**Applicable Preference Documentation URL(s)**: https://support.zoom.us/hc/en-us/articles/10365151862285; https://support.zoom.us/hc/en-us/articles/115001799006

apizz:
Closes #583 

- Add preferences to be in alignment with documentation
- Add auto update v2 preferences
- Update menus to be in alignment with documentation structure

### DIFF
--- a/Manifests/ManagedPreferencesApplications/us.zoom.config.plist
+++ b/Manifests/ManagedPreferencesApplications/us.zoom.config.plist
@@ -142,6 +142,17 @@
 			<dict>
 				<key>Install and update</key>
 				<array>
+					<string>AU2_EnableAutoUpdate</string>
+					<string>AU2_DeploySpecificVersion</string>
+					<string>AU2_EnableManualUpdate</string>
+					<string>AU2_EnablePromptUpdateForAU2</string>
+					<string>AU2_EnableShowZoomUpdates</string>
+					<string>AU2_EnableUpdateAvailableBanner</string>
+					<string>AU2_EnableUpdateSuccessNotification</string>
+					<string>AU2_InstallAtIdleTime</string>
+					<string>AU2_SetUpdateChannel</string>
+					<string>AU2_SafeUpgradePeriod</string>
+					<string>AU2_SetUpdateSchedule</string>
 					<string>DisableDaemonInstall</string>
 					<string>SetEnrollToken4CloudMDM</string>
 					<string>SetUpdatingChannel</string>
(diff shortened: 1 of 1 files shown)


kevinmcox (approved):
Looks OK to me. It loads fine in PC with no issues, but I did not export and test every setting.

Manifests/ManagedPreferencesApplications/us.zoom.config.plist
```diff
@@ -119,128 +119,237 @@
...
+				<string>Screen Sharing</string>
+				<string>Zoom Phone</string>
+				<string>Meeting Chat</string>
+				<string>Apps</string>
+				<string>Background / Filters</string>
+				<string>Recording</string>
+				<string>Zoom Room</string>
 				<string>Chat</string>
+				<string>Misc.</string>
+				<string>ZfG</string>
```
relgit, line 137: Maybe let's change this to _ZfG Cloud_?

Manifests/ManagedPreferencesApplications/us.zoom.config.plist
```diff
@@ -284,10 +393,12 @@
 			<string>SSO URL</string>
 			<key>pfm_type</key>
 			<string>string</string>
+			<key>pfm_value_placeholder</key>
+			<string>zoom_subdomain</string>
```
relgit, line 397: I didn't immediately understand "zoom_subdomain". Maybe let's change it to "Example: 'myorg' when using a login URL of myorg.zoom.us"?

I think we have precedence in the repository of placeholders starting with "Example:".

Manifests/ManagedPreferencesApplications/us.zoom.config.plist
```diff
@@ -312,6 +423,8 @@
 			<string>Lock SSO URL</string>
 			<key>pfm_type</key>
 			<string>string</string>
+			<key>pfm_value_placeholder</key>
+			<string>zoom_subdomain</string>
```
relgit, line 427: Same as above.

Manifests/ManagedPreferencesApplications/us.zoom.config.plist
```diff
@@ -483,6 +598,402 @@
...
+			<string>Set the cadence of updates applied to the desktop client</string>
+			<key>pfm_documentation_url</key>
+			<string>https://support.zoom.us/hc/en-us/articles/10365151862285</string>
+			<key>pfm_exclude</key>
+			<array>
+				<dict>
+					<key>pfm_target_conditions</key>
+					<array>
+						<dict>
+							<key>pfm_present</key>
```
relgit, line 634: I realize this block was taken from preferences already existing in the manifest, but I think the two conditions contradict.
I would replace the two conditions:
```
<key>pfm_present</key>
<false/>
<key>pfm_range_list</key>
<array>
	<false/>
</array>
```
With this one:
```
<key>pfm_n_range_list</key>
<array>
	<true/>
</array>
```

relgit (changes requested):
A lot of work there @apizz, and the majority is superb. I just have a few comments.

Manifests/ManagedPreferencesApplications/us.zoom.config.plist
apizz: I'm happy to do this for other preferences where the `pfm_default` = `false`, but my hesitation with doing this with the AU2 settings is because `AU2_EnableAutoUpdate` defaults to `true`.  As a result, even if set to the way you propose it could produce a situation where even if it's not actually added to a profile it is still treated as valid because the default value is `true` if that makes sense.  So for those AU2 settings, I'd like to leave the `pfm_present` keys in place, just in case

relgit: In this case, instead of using what I suggested we should just remove the presence test and keep the range list as it is, like so:
```
<array>
    <dict>
        <key>pfm_range_list</key>
        <array>
        	<false/>
        </array>
        <key>pfm_target</key>
        <string>AU2_EnableAutoUpdate</string>
    </dict>
</array>
```
This way, `AU2_SetUpdateChannel` and the like will only be excluded when `AU2_EnableAutoUpdate` is set to `false`. When the latter is not present, it is not set to false and therefore the former keys will be enabled.

I've tested it in both IPE and PC and it's working well.

relgit (approved):

PR #589, opened 2022-12-23
New Manifest - Twingate macOS Client

briangoldstein:
Adding a manifest for the Twingate macOS Client. There isn't much to control yet, but I think this will be helpful since two of these keys are currently undocumented and not everyone is comfortable with mcx to profile conversions.


Manifests/ManagedPreferencesApplications/com.twingate.macos.plist
```diff
@@ -0,0 +1,192 @@
...
+			<key>pfm_description</key>
+			<string>Suppresses the first run Data Privacy screen</string>
+			<key>pfm_name</key>
+			<string>PresentedDataPrivacy</string>
+			<key>pfm_title</key>
+			<string>Suppress Data Privacy Screen</string>
+			<key>pfm_type</key>
+			<string>boolean</string>
+			<key>pfm_documentation_url</key>
+			<string>https://www.twingate.com/privacy/</string>
```
kevinmcox, line 176: I don't see anything at this URL related to the `PresentedDataPrivacy` key. My thought would be to remove the URL if it isn't specifically taking us to documentation of the key.

Manifests/ManagedPreferencesApplications/com.twingate.macos.plist
```diff
@@ -0,0 +1,192 @@
...
+			<key>pfm_name</key>
+			<string>PayloadOrganization</string>
+			<key>pfm_title</key>
+			<string>Payload Organization</string>
+			<key>pfm_type</key>
+			<string>string</string>
+		</dict>
+		<dict>
+			<key>pfm_description</key>
+			<string>Sets the network address for first run</string>
```
kevinmcox, line 140:
```suggestion
			<string>A pre-configured Twingate network subdomain (eg. autoco from autoco.twingate.com). This avoids users needing to enter your Twingate network address on first run.</string>
```

I'd be tempted to use this more descriptive language direct from the documentation link.

Manifests/ManagedPreferencesApplications/com.twingate.macos.plist
```diff
@@ -0,0 +1,192 @@
...
+			<key>pfm_name</key>
+			<string>network</string>
+			<key>pfm_title</key>
+			<string>Network Subdomain</string>
+			<key>pfm_type</key>
+			<string>string</string>
+			<key>pfm_value_placeholder</key>
+			<string>autoco</string>
+			<key>pfm_documentation_url</key>
```
kevinmcox, line 149: Can this be sorted alphabetically please?

Manifests/ManagedPreferencesApplications/com.twingate.macos.plist
```diff
@@ -0,0 +1,192 @@
...
+			<string>https://www.twingate.com/docs/macos-and-ios/#twingate-network-pre-configuration</string>
+		</dict>
+		<dict>
```
kevinmcox, line 152: Can we add the same [documentation URL](https://www.twingate.com/docs/macos-and-ios/#twingate-network-pre-configuration) to this key.

kevinmcox (changes requested):

Manifests/ManagedPreferencesApplications/com.twingate.macos.plist
briangoldstein: My thinking was to link the admin to the privacy policy since they are skipping/agreeing it on behalf of their users, but I see how that can be overkill and not add direct value for the admin creating the profile. Will remove.

Manifests/ManagedPreferencesApplications/com.twingate.macos.plist
briangoldstein: Thanks for the suggestion @kevinmcox, I was struggling with this one, and your solution makes perfect sense.

kevinmcox (approved):
Changes look good to me, thanks for the contribution!

apizz:
Ran through `plutil` to cleanup some extra whitespace and preference ordering

apizz:
@briangoldstein the only thing I would say here is it would be good to define what versions support the following preferences.  This way as more preferences are added over time there's no confusion about what versions support which preferences of the app.

The only other nit picky thing I have is the root-level description indicates only for macOS but presumably iOS support _could_ come at some future date.  While not important now since the profile is only macOS, if it is extended this means we both would need to:
1. Update the description
2. Specify for each preference whether it's supported for both iOS & macOS or just one or the other.

I consider both above not essential, but if we _did_ do it now means less we have to remember to change or account for later.

briangoldstein:
@apizz Thanks for the feedback! I have added the **pfm_app_min** key to the four current preferences.  
The preference domain is **com.twingate.macos**, which I assume would not be used for their iOS app, but I'm happy to make the OS related changes if you think it would still be worth it.

apizz (approved):

PR #598, opened 2023-02-01
Enhance 1Password 8 manifest with iOS keys

fspeirs:
Sorry about the earlier botched PR. This one should be more easily mergeable - it adds information about the keys supported by iOS and clarifies some of the text.

Is it OK to have an overall `pfm_app_min` at the top of the file, rather than having them in each item, given that all of these keys are supported on 8.9 and later on both platforms?

Also, I couldn't figure out how to have a different `pfm_range_min` value for different platforms for the key `security.autolock.minutes`. It should be 1 on macOS and 0 on iOS. I'm happy to adapt this PR if you can advise.

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.1password.1password.plist
+++ b/Manifests/ManagedPreferencesApplications/com.1password.1password.plist
@@ -3,7 +3,7 @@
 <plist version="1.0">
 <dict>
 	<key>pfm_app_min</key>
-	<string>8.0</string>
+	<string>8.9</string>
 	<key>pfm_app_url</key>
 	<string>https://1password.com</string>
 	<key>pfm_description</key>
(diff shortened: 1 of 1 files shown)


apizz:
@fspeirs The idea is that the top-level `pfm_app_min` version in the manifest should reflect "high-level" minimum app version that supported management.  As new preferences are added or are supported on different platforms, that's when you'd use the `pfm_app_min` _at the preference level_ to indicate what's supported.  That said, gaining iOS support for a set of preferences with `pfm_app_min` isn't possible on a per-OS basis, so I think the following is best:

1. Leave top-level `pfm_app_min` at `8.0`.
2. For any existing preferences that were supported on macOS that are now also supported on iOS, leave the current `pfm_app_min` specified in the preference and add a `pfm_note` indicating what app version (8.9?) gained iOS support.
3. For new prefs that are only supported on iOS, add `pfm_app_min` at `8.9` for each of those preferences.

apizz (changes requested):
Per prev. comment

kevinmcox:
Closing due to a lack of activity.

PR #602, opened 2023-02-24
Add CrossSiteTrackingPreventionRelaxedDomains

WardsParadox:
Adds new keys from 13.1, 16.2

https://developer.apple.com/documentation/devicemanagement/domains

### DIFF
--- a/Manifests/ManifestsApple/com.apple.domains.plist
+++ b/Manifests/ManifestsApple/com.apple.domains.plist
@@ -295,6 +295,6 @@ SafariPasswordAutoFillDomains definitions are cumulative. Patterns defined by al
 	<key>pfm_unique</key>
 	<true/>
 	<key>pfm_version</key>
-	<integer>1</integer>
+	<integer>2</integer>
 </dict>
 </plist>
--- a/Manifests/ManifestsApple/com.apple.domains.plist
+++ b/Manifests/ManifestsApple/com.apple.domains.plist
@@ -13,7 +13,7 @@
 	<key>pfm_interaction</key>
 	<string>combined</string>
 	<key>pfm_last_modified</key>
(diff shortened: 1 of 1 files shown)


Manifests/ManifestsApple/com.apple.domains.plist
```diff
@@ -246,6 +246,48 @@ SafariPasswordAutoFillDomains definitions are cumulative. Patterns defined by al
...
+			</array>
+			<key>pfm_repetition_max</key>
+			<integer>10</integer>
+			<key>pfm_subkeys</key>
+			<array>
+				<dict>
+					<key>pfm_description</key>
+					<string></string>
+					<key>pfm_name</key>
+					<string>CrossSiteTrackingDomain</string>
```
apizz, line 275: If this is indeed an array of strings, `pfm_name` should be removed here.  `pfm_title` is the only thing you need besides the `pfm_type` and `pfm_value_placeholder`.  Remove the empty `pfm_description`

apizz (changes requested):
Anytime you contribute, you should update the top-level `pfm_last_modified` date value and in this case where you're adding new preferences the `pfm_version` at the very bottom should also be bumped.

Other than that and the comment I made lgtm

Manifests/ManifestsApple/com.apple.domains.plist
WardsParadox: Updating now

WardsParadox: Updated in 48cd1a0

relgit (approved):
Thanks for making the changes @WardsParadox.

apizz:
Thanks again

apizz (approved):

PR #611, opened 2023-03-10
Fix 'github-actions[bot]' to point to correct GitHub App

zacharyfleck:
Simple change to make commits match up to the official 'github-actions' app in GitHub, rather than an unmatched user.

### DIFF
--- a/.github/workflows/updateIndexes.yml
+++ b/.github/workflows/updateIndexes.yml
@@ -20,8 +20,8 @@ jobs:
         working-directory: clone
         run: |
           ./updateIndexes.py
-          git config user.name git config user.name "github-actions[bot]"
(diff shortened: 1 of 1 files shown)


zacharyfleck:
> Look like a redundant paste sneaked into line 23.

Whoops, that's fixed.

> Also, is there documentation about this email address on line 24 posted anywhere? I couldn't find, although I didn't dig too deep.

Official docs, I don't really have. Only can reference other users' success from the forums. This one from a while back states that the numbers aren't required:

1. https://github.com/actions/checkout/discussions/479

Meanwhile, this more recent post suggest they are:

2.  https://github.com/orgs/community/discussions/26560

I can say with confidence that I have used the email address with the preceding numbers with success in our org's repositories. I have updated the PR without the numbers, but haven't had a moment to test it yet. Your feedback is appreciated.

PR #614, opened 2023-03-29
Added Manifest V2 Controls for Google Chrome

northportio:


apizz:
Thanks for your contribution.  However, it does not appear you fully read the contribution guide as you did not include a pfm_version update or a `pfm_last_modified` update in this manifest.  There were also some errors which I've corrected.

In future, please don't include large aesthetic changes (escaping quotes and single apostrophes, which aren't actually necessary in xml) in a minor functional change as it just makes it harder to parse what's actually changed.

PR #623, opened 2023-07-10
Add profile manifests for Alectrona Patch

ryangball:
This PR is to add Profile Manifests for Alectrona Patch.


apizz (changes requested):
@ryangball thank you for your contribution.  However, please clean up your file such that:
1. there are no empty blank lines (there are many)
2. key/value pairs are sorted alphabetically
3. XML spacing is properly aligned (there's a number of `<dict>` tags that aren't properly aligned)

`plutil` can assist with much of that.

relgit (approved):
Thanks @ryangball 🙂

PR #641, opened 2023-11-19
Fix pfm_format_version in com.vpntracker.365mac-config.plist

homebysix:
The only currently valid value for `pfm_format_version` is `<integer>1</integer>`.

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.vpntracker.365mac-config.plist
+++ b/Manifests/ManagedPreferencesApplications/com.vpntracker.365mac-config.plist
@@ -11,7 +11,7 @@
 	<key>pfm_format_version</key>
 	<integer>1</integer>
 	<key>pfm_last_modified</key>
-	<date>2022-11-11T09:00:00Z</date>
+	<date>2023-11-19T21:41:10Z</date>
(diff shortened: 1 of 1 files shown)


Manifests/ManagedPreferencesApplications/com.vpntracker.365mac-config.plist
```diff
@@ -234,6 +234,6 @@ A profile can consist of payloads with different version numbers. For example, c
 	<key>pfm_unique</key>
 	<true/>
 	<key>pfm_version</key>
-	<integer>1</integer>
+	<integer>2</integer>
```
relgit, line 237: The version number can remain 1 (by convention it is only increased when preference keys are added or removed).

relgit (changes requested):
Thanks! See my comment.

Manifests/ManagedPreferencesApplications/com.vpntracker.365mac-config.plist
homebysix: Good to know. Fixed.

relgit (approved):

PR #642, opened 2023-11-19
Add missing documentation URLs to com.microsoft.OneDrive.plist

homebysix:
All missing URLs are anchors on this page: https://learn.microsoft.com/en-us/sharepoint/deploy-and-configure-on-macos


homebysix:
Base wasn't up to date — disregard.

PR #645, opened 2023-11-25
Digi dna official descriptions REBASE

apizz:
Rebase of #626

PR #626, opened 2023-08-21
Official descriptions

relgit:
This PR updates all `pfm_description` tags throughout the active manifests with official text from Apple's Device Management repository. For system manifests, this applies to all keys where descriptions are available. For other manifests this is limited to common keys only (e.g. `PayloadType`).

Having up-to-date official description texts means that `pfm_description_reference` tags are no longer necessary in changed system manifests, so I have removed them from those manifests.

To avoid future merge conflicts, this PR if based off of #609, which is why the changes seem enormous for now. Once #609 is merged, we can come back here to review the actual changes and merge.

PR #648, opened 2023-11-30
Official descriptions rerun

relgit:
This is a rerun of the scripts that created #626, off of the current state of the repository. It is meant to replace both #626 and #645, which have become full of unrelated commits.

To recap, in this PR:
* System manifest preference key descriptions updated with official texts where available, except for root descriptions where the official ones are empty
* Common preference keys in all manifests as well as in the manifest template updated with official descriptions
* Reference description tags removed from updated preference keys


relgit:
I think that this PR has become too stale to merge at this point. If there's still interest in updating to the official descriptions, I'm happy to generate a new one after the current OS beta period.

PR #649, opened 2023-12-12
Update manifests for Tailscale 1.56

agottardo:
This PR updates the manifests for Tailscale (App Store and Standalone variants of the client).

- Adds 5 additional settings that are being released in Tailscale 1.56 and can be managed via MDM.
- Fixes a typo.


apizz:
We need to increment the manifest version and update the pfm_last_modified date but otherwise lgtm

agottardo:
@apizz updated! Thanks.

PR #652, opened 2024-01-17
Update com.secondsonconsulting.baseline.plist

BigMacAdmin:
Updates to support new features of [Baseline](https://github.com/secondsonconsulting/Baseline) project.

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.secondsonconsulting.baseline.plist
+++ b/Manifests/ManagedPreferencesApplications/com.secondsonconsulting.baseline.plist
@@ -13,7 +13,7 @@
 	<key>pfm_format_version</key>
 	<integer>1</integer>
 	<key>pfm_last_modified</key>
-	<date>2024-01-16T00:00:00Z</date>
+	<date>2024-01-24T00:00:00Z</date>
 	<key>pfm_platforms</key>
 	<array>
 		<string>macOS</string>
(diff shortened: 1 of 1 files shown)


relgit (changes requested):
Thanks for the submission B.

On boolean properties with default value, we prefer using `pfm_default` to indicate the default over mentioning it in `pfm_description`. 

GUI consumers use this to correctly tick checkboxes as preview and in some cases to decide if properties should be written to a profile.

And, another thing, please bump up the version number in `pfm_version`.

Cheers!

BigMacAdmin:
> On boolean properties with default value, we prefer using `pfm_default` to indicate the default over mentioning it in `pfm_description`.

Thanks for the quick reply!

If I use `pfm_default` will iMazing or ProfileCreator actually write that key to every profile? Or will it omit that key, unless the user chooses to explicitly set it?

Is `pfm_default` a "free form" type key where I type in plain language, or do i need to set it as if it were a boolean value itself?

relgit:
> If I use pfm_default will iMazing or ProfileCreator actually write that key to every profile? Or will it omit that key, unless the user chooses to explicitly set it?

The latter. Both editors use this to provide a visual cue to users so that telling the default value of a preference is a matter of a quick glance at the checkbox. iMazing Profile Editor also uses this to omit boolean keys when they are set to their default values unless forced by the user (Option+click) or configured not to.

> Is pfm_default a "free form" type key where I type in plain language, or do i need to set it as if it were a boolean value itself?

The expected type is the same as the type of the property. In other words, for boolean properties the value must be either `<true/>` or `<false/>`.

BigMacAdmin:
> > If I use pfm_default will iMazing or ProfileCreator actually write that key to every profile? Or will it omit that key, unless the user chooses to explicitly set it?
> 
> The latter. Both editors use this to provide a visual cue to users so that telling the default value of a preference is a matter of a quick glance at the checkbox. iMazing Profile Editor also uses this to omit boolean keys when they are set to their default values unless forced by the user (Option+click) or configured not to.
> 
> > Is pfm_default a "free form" type key where I type in plain language, or do i need to set it as if it were a boolean value itself?
> 
> The expected type is the same as the type of the property. In other words, for boolean properties the value must be either `<true/>` or `<false/>`.

Done and done, thanks again.

relgit (approved):
Thanks!

PR #655, opened 2024-01-30

doschupp:
With version 3 of the Network Share Mounter, the configuration has changed a little bit and new configuration values are also available.

### DIFF
--- a/Manifests/ManagedPreferencesApplications/de.fau.rrze.NetworkShareMounter.plist
+++ b/Manifests/ManagedPreferencesApplications/de.fau.rrze.NetworkShareMounter.plist
@@ -11,9 +11,9 @@
 	<key>pfm_domain</key>
 	<string>de.fau.rrze.NetworkShareMounter</string>
 	<key>pfm_format_version</key>
-	<integer>2</integer>
+	<integer>1</integer>
 	<key>pfm_last_modified</key>
-	<date>2024-01-30T09:30:00Z</date>
+	<date>2024-01-31T13:04:00Z</date>
 	<key>pfm_platforms</key>
 	<array>
 		<string>macOS</string>
@@ -277,6 +277,6 @@ A profile can consist of payloads with different version numbers. For example, c
 	<key>pfm_unique</key>
 	<true/>
 	<key>pfm_version</key>
-	<integer>1</integer>
+	<integer>2</integer>
 </dict>
 </plist>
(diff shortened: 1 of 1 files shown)


kevinmcox (changes requested):
Thanks for the contribution @doschupp.

Please not the pre-commit error:

> Manifests/ManagedPreferencesApplications/de.fau.rrze.NetworkShareMounter.plist: pfm_format_version should be 1, not 2 (https://github.com/ProfileCreator/ProfileManifests/wiki/Manifest-Format-Versions)

Line 14 needs to revert back to `1` while Line 280 needs to be bumped to `2`.

Manifests/ManagedPreferencesApplications/de.fau.rrze.NetworkShareMounter.plist
```diff
@@ -150,24 +150,119 @@ A profile can consist of payloads with different version numbers. For example, c
...
+					<string>^smb://\S*/.+$</string>
 					<key>pfm_type</key>
 					<string>string</string>
 					<key>pfm_value_placeholder</key>
 					<string>smb://filer.your.domain/share</string>
 				</dict>
 			</array>
 			<key>pfm_title</key>
-			<string>List of Network Shares to Mount</string>
+			<string>Network shares (Deprecated)</string>
```
relgit, line 265: Instead of changing the title to reflect deprecation, could you add the `pfm_app_deprecated` tag to the subkey?

Manifests/ManagedPreferencesApplications/de.fau.rrze.NetworkShareMounter.plist
```diff
@@ -150,24 +150,119 @@ A profile can consist of payloads with different version numbers. For example, c
 			<string>boolean</string>
 		</dict>
 		<dict>
+			<key>pfm_default</key>
```
relgit, line 153: Please add the `pfm_app_min` to all keys introduced with version 2.

relgit (changes requested):

Manifests/ManagedPreferencesApplications/de.fau.rrze.NetworkShareMounter.plist
relgit: My bad, I meant to say version 3, though I see that you did that too. Thanks!

relgit:
I see that the description of the network shares property is also reflecting the deprecation, like the title did, so we need that reverted as well. I'll quickly do that myself so we can merge this with no further delays.

PR #659, opened 2024-03-07
Update com.barebones.bbedit.plist

RedClawx:
Added all available preferences into the manifest for BBEdit 15.0.1, including all the Expert preference settings.

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.barebones.bbedit.plist
+++ b/Manifests/ManagedPreferencesApplications/com.barebones.bbedit.plist
@@ -7,13 +7,13 @@
 	<key>pfm_description</key>
 	<string>BBEdit settings</string>
 	<key>pfm_documentation</key>
-	<string>https://www.barebones.com/support/bbedit/ExpertPreferences.html</string>
+	<string>https://www.barebones.com/support/bbedit/</string>
 	<key>pfm_domain</key>
 	<string>com.barebones.bbedit</string>
 	<key>pfm_format_version</key>
 	<integer>1</integer>
 	<key>pfm_last_modified</key>
-	<date>2023-08-17T13:20:39Z</date>
+	<date>2024-02-28T13:44:00Z</date>
 	<key>pfm_platforms</key>
 	<array>
 		<string>macOS</string>
@@ -24,7 +24,7 @@
 			<key>pfm_default</key>
 			<string>Configures BBEdit settings</string>
 			<key>pfm_description</key>
-			<string>Description of the payload</string>
+			<string>The human-readable description of this payload. This description appears on the Detail screen.</string>
 			<key>pfm_name</key>
 			<string>PayloadDescription</string>
 			<key>pfm_title</key>
@@ -36,7 +36,7 @@
 			<key>pfm_default</key>
 			<string>BBEdit</string>
 			<key>pfm_description</key>
-			<string>Name of the payload</string>
+			<string>The human-readable name for the profile payload. The name appears on the Detail screen and doesn't need to be unique.</string>
 			<key>pfm_name</key>
 			<string>PayloadDisplayName</string>
 			<key>pfm_require</key>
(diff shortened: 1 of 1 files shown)


relgit (changes requested):
This is an incredible amount of work, thank you @RedClawx.

I can see some iMazing-only PFM tags, so I'm assuming it must have been tested on iMazing Profile Editor. Were you able to test on ProfileCreator?

I found one property that has been completely removed, `AlwaysUseSystemAppearanceForTitleBar`. We prefer keeping deprecated and removed properties and marking them accordingly (`pfm_app_deprecated` and `pfm_app_max`).

Could you please restore that property and others that may have been removed?

RedClawx:
Good day,
I was testing specifically with iMazing Profile Editor. I did not do any testing with ProfileCreator.

For the removed keys: The reason I removed tham was due to them no longer being available from BBEdit 12.6.7 or better. version 12.6.7 was as far back as I could go. I'm not able to run anything earlier. The OS I was running for compiling all the keys was macOS 13. Any keys I wasn't able to locate were removed as I don't really know when those kays had been removed from BBEdit. The only versions of BBEdit I gathered keys from were from 12.6.7 through 15.0.1.

I can resotre the keys from the original entries, but I wouldn't know at what point they were deprecated and removed.

RedClawx:
The follow keys have been removed or replaced. If you still want me to add the keys back in, is it okay to add them under to a seperate tab called "Removed"?

0. _Old Key_
   - _Replaced by_
1. AlwaysUseSystemAppearanceForTitleBar
   - ApplicationAppearanceMode
2. GitToolPathOverride
3. MoveModelessWindowsToActiveSpace
   - MoveModelessWindowToActiveSpace
4. MoveModelessWindowsToActiveSpace_FindWindow
5. MoveModelessWindowsToActiveSpace_MultiFileFindWindow
6. MoveModelessWindowsToActiveSpace_OpenFileByNameWindow
7. NSRecentDocumentsLimit:Folders
   - NSRecentDocumentsLimit_Folders
8. NSRecentDocumentsLimit:Images
   - NSRecentDocumentsLimit_Images
9. NSRecentDocumentsLimit:Projects
   - NSRecentDocumentsLimit_Projects
10. QuickTimeImages
11. ReopenRemoteDocuments
    - ReopenDocumentsOnRemoteVolumes
12. SUAllowsAutomaticUpdates
    - SUSoftwareUpdateEnabled
13. SUAutomaticallyUpdate
    - SUSoftwareUpdateEnabled
14. SubversionToolPathOverride
15. SUEnableAutomaticChecks
    - SUSoftwareUpdateEnabled
16. SUFeedURL
17. SUScheduledCheckInterval
18. UseSystemAppearanceWhenInDarkMode
    - ApplicationAppearanceMode

relgit:
Thanks for detailing your method.

Several keys have indeed been removed but I'd like to keep them. Others were simply never documented and should also remain.

Here's a breakdown of which ones should remain and why:

All keys beginning with `MoveModelessWindowsToActiveSpace` were removed recently from the expert preference page, so let's keep and tag as deprecated in 15.0. Also, I can't find the replacement you mentioned (`MoveModelessWindowToActiveSpace`) in the updated manifest.

`QuickTimeImages` was removed from version 15.0 so let's keep and tag accordingly.

`ReopenRemoteDocuments` same as above.

All the keys beginning with `SU` belong to the Sparkle update library. Those are usually undocumented but should definitely remain. Also, were did you find the replacement key `SUSoftwareUpdateEnabled`? Looking at the user manual doesn't reveal the actual key name.

I also can't find the new `ApplicationAppearanceMode` in the documentation - how did you arrive at the key's name?

That's it. The rest can stay removed. After you restore the above and clarify the method for arriving at the new key names we'd be able to merge the work. Thanks!

RedClawx:
I made the changed requested. I added 2 tabs for them. The first tab is for the Removed Expert keys that are no longer in version 15. The second tab is for the "SU" keys.

`MoveModelessWindowToActiveSpace` was also added in. I must have skiped over that key when building the new manifest.

`SUSoftwareUpdateEnabled` and `ApplicationAppearanceMode`, as with all the other keys I collected, was found by comparing the preference file for BBEdit after making a change to the preferences. I copied the BBedit pref once to get the default settings. Then I launched BBEdit, made a change to its settings. Then used BBEdit on another computer to compare the changed preference file with the unchanged preference file. It was a pain oin the rear to do, but it did get me the actual key names for almost every setting.

PR #666, opened 2024-03-26
DigiDNA Mar 2024

relgit:
This PR includes a collection a small fixes and updates:

* Incorrect default values removed and new property added in the Firefox manifest
* Payload descriptions updated to convention on a number of application manifests
* New property added to the iOS Restriction manifest
* Outdated note removed from the macOS and tvOS Restrictions manifests
* Minimum macOS version corrected on a number of system manifests
* Property updated to reflect documentation corrections in the DNS Settings manifest
* Property value type corrected in the Education manifest
* Two properties added and several properties augmented in the Extensible SSO manifest
* The Network Usage Rules manifest is no longer marked as requiring supervision
* Property changed to reflect the most recent documentation in the Web Clip manifest
* One property added, incorrect value type corrected, and multiple conditionals added in the Wi-Fi manifest

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.1password.1password.plist
+++ b/Manifests/ManagedPreferencesApplications/com.1password.1password.plist
@@ -7,7 +7,7 @@
 	<key>pfm_app_url</key>
 	<string>https://1password.com</string>
 	<key>pfm_description</key>
-	<string>1Password 8 settings</string>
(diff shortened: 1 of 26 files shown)


relgit:
Addendum:
* Two missing keys added to the Wi-Fi manifest

PR #668, opened 2024-04-23
Addressing method deprecation in the updateIndexes.py script

relgit:
Beginning with Python 3.12, our index update script is showing a deprecation warning. This PR replaces the deprecated method call with its modern equivalent.


apizz (approved):
Lgtm. Ty!

PR #671, opened 2024-05-13
Baseline2.2 updates

BigMacAdmin:
- Updating for Baseline v2.2 features
- Improved descriptions for some existing features
- Updated the last modified date
- Updated the pfm_version
- Updated to new Baseline icon


Manifests/ManagedPreferencesApplications/com.secondsonconsulting.baseline.plist
```diff
@@ -431,6 +431,82 @@ A profile can consist of payloads with different version numbers. For example, c
...
+							<key>pfm_name</key>
+							<string>Icon</string>
+							<key>pfm_title</key>
+							<string>Icon</string>
+							<key>pfm_type</key>
+							<string>string</string>
+						</dict>
+						<dict>
+							<key>pfm_app_min</key>
+							<string>2.1</string>
```
relgit, line 478: Likely sneaked in by mistake, seeing that the container is 2.2 minimum.

Manifests/ManagedPreferencesApplications/com.secondsonconsulting.baseline.plist
```diff
@@ -431,6 +431,82 @@ A profile can consist of payloads with different version numbers. For example, c
...
+			<key>pfm_title</key>
+			<string>WaitFor</string>
+			<key>pfm_type</key>
+			<string>array</string>
+		</dict>
+		<dict>
+			<key>pfm_app_min</key>
+			<string>2.2</string>
+			<key>pfm_description</key>
+			<string>When using WaitFor, enter the number of seconds you wish to wait before Baseline considers remaining items a failure. This timer does not start until all other items are processed. Default 300.</string>
```
relgit, line 502: If you replace the default value mention in the description with a `pfm_default` element you'll get a better presentation.

relgit (commented):
Hey @BigMacAdmin, thanks for the update.

See my two easily addressable comments. Cheers.

Manifests/ManagedPreferencesApplications/com.secondsonconsulting.baseline.plist
BigMacAdmin: That 2.1 is actually on purpose, back filling something I missed with my last update.

Manifests/ManagedPreferencesApplications/com.secondsonconsulting.baseline.plist
BigMacAdmin: Done!

Manifests/ManagedPreferencesApplications/com.secondsonconsulting.baseline.plist
relgit: If that's the case then the container, `WaitFor`, was already supported in 2.1 and therefore should declare it as minimum instead of 2.2. Then assuming that `DisplayName`, `Path`, and `Icon` are new, they could declare their minimum as 2.2.

Although that's how it should be marked up if I understood you correctly, I suspect there's still some detail missing because if I try to understand the role of `WaitFor`, it doesn't really make sense to me that it was introduced in 2.1 with only `Subtitle`.

BigMacAdmin: Apologies for the confusion, you are correct in your original assessment that I had the wrong pfm_app_min for the `WaitFor` . I misread which key we were discussing. I've fixed this, thank you as always for your attention to detail.

relgit (approved):

PR #674, opened 2024-05-21
Update com.apple.extensiblesso.plist

bgatestmg:
This should fix the crashing issue for this payload. at least in my localized testing it has.

### DIFF
--- a/Manifests/ManifestsApple/com.apple.extensiblesso.plist
+++ b/Manifests/ManifestsApple/com.apple.extensiblesso.plist
@@ -3,9 +3,7 @@
 <plist version="1.0">
 <dict>
 	<key>pfm_description</key>
-	<string>Single Sign-On Extension settings. User-level payload support for macOS 11.x and later.</string>
-	<key>pfm_description_reference</key>
-	<string>The payload for configuring an app extension that performs single sign-on.</string>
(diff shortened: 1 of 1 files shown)


homebysix:
Please make this update upstream. See the readme for details.

kevinmcox (approved):
Confirmed this fixes the crash with ProfilesCreator 0.3.3

Manifests/ManifestsApple/com.apple.extensiblesso.plist
```diff
@@ -2149,5 +2235,55 @@ kerberosDefault - The default Kerberos processes for selecting credentials is us
 	<true/>
 	<key>pfm_version</key>
 	<integer>6</integer>
```
kevinmcox, line 2237: Can you bump the `pfm_version` to 7 please.

kevinmcox (changes requested):

relgit:
Isn't this an as-is copy of the DigiDNA fork manifest?

I'm glad it works and we've never kept our manifest work to ourselves, but there are too many changes to tell us what actually makes the difference.

Could you perhaps do some elimination?

kevinmcox:
The latest release of Profile Creator no longer crashes when this payload is loaded.

PR #676, opened 2024-05-28
Add support companion manifest

almenscorner:
Adding manifest for https://github.com/macadmins/supportcompanion


relgit (changes requested):
Thanks for the submission @almenscorner.

Please see my three comments.

When you're done, please also run the file through `plutil -convert xml1 [file name]` so it matches the repository conventions.

Also, would you like to add the app icon too? You can simply add a `.png` file named using the app bundle ID to the `/Icons/ManagedPreferencesApplications/` directory.

almenscorner:
Thank you for the suggestions, I have updated with your suggestions and also added an Icon!

relgit (approved):
Thanks for making the changes! Looks like running through plutil was skipped, but I can sort it out.

PR #684, opened 2024-06-24
Add new desktop info configurations

almenscorner:
This update adds support for the new desktop info configurations introduced in Support Companion 1.1.0


Manifests/ManagedPreferencesApplications/com.almenscorner.supportcompanion.plist
```diff
@@ -530,6 +542,163 @@
...
+			<key>pfm_name</key>
+			<string>DesktopInfoLevel</string>
+			<key>pfm_range_list</key>
+			<array>
+				<string>Full</string>
+				<string>Minimal</string>
+				<string>Hardware</string>
+				<string>Custom</string>
+			</array>
+			<key>pfm_range_list_titles</key>
```
relgit, line 595: `pfm_range_list_titles` is unnecessary here if the raw values are identical.

Manifests/ManagedPreferencesApplications/com.almenscorner.supportcompanion.plist
```diff
@@ -9,11 +9,11 @@
 	<key>pfm_documentation_url</key>
 	<string>https://github.com/macadmins/supportcompanion</string>
 	<key>pfm_domain</key>
-	<string>com.almenscorner.supportcompanion</string>
+	<string>SupportCompanion</string>
```
relgit, line 12: `pfm_domain` must be equal to the bundle ID of the app (and so is `pfm_default` of the `PayloadType` key, by the way).

almenscorner: The reason I changed this to "SupportCompanion" is because the app is looking for the domain "SupportCompanion" when getting the settings from the mobileconfig. Changing the PayloadType to the bundle id breaks this behaviour and leaving this key as the bundle id breaks opening up an already created profile in iMazing for example, then you get a "Unfammiliar domain" error.

Manifests/ManagedPreferencesApplications/com.almenscorner.supportcompanion.plist
almenscorner: I have removed the list titles

Manifests/ManagedPreferencesApplications/com.almenscorner.supportcompanion.plist
relgit: I see, your app is reading from a different suite name and not from `UserDefaults.standard`. It's rare that I've seen that, but it exists and will work.

So just to clarify what I said previously, `pfm_domain` must be equal to the _preference domain_ that the app is reading from. Usually that's the app bundle ID 🙂

The unfamiliar domain notice is a side effect of `pfm_domain` and the `pfm_default` of `PayloadType` not being equal, so you've fixed that one indeed.

Last thing I should probably mention is that it is a bit peculiar that the preferences domain is not a reverse-DNS representation of a domain name that you own. Not a big deal, the namespace is probably wide open, just a bit unconventional. 

However, we do usually ask that a manifest file name match the domain name. In this case this means that the file should be renamed _SupportCompanion.plist_ which does stand out a bit. 

@apizz, @kevinmcox – what do you think?

relgit: @almenscorner, looks like AP and Kevin have left this for me to handle so let's just rename the manifest to the preferences domain name (SupportCompanion.plist) and then I'll merge it.

kevinmcox: Sorry I'm just seeing this.

I agree it is a little unusual, but I think a good comparison is the manifest for Munki which is just `ManagedInstalls` in both preference domain and filename:

https://github.com/ProfileCreator/ProfileManifests/blob/master/Manifests/ManagedPreferencesApplications/ManagedInstalls.plist

So renaming the file at least keeps us consistent with existing manifests.

almenscorner: They have now been renamed, let me know if anything else is needed from my end

kevinmcox (approved):

PR #697, opened 2024-09-03
profile manifest for XCreds 5.0 (7176)

twocanoes:
profile manifest for XCreds 5.0 (7176)

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.twocanoes.xcreds.plist
+++ b/Manifests/ManagedPreferencesApplications/com.twocanoes.xcreds.plist
@@ -13,7 +13,7 @@
 	<key>pfm_format_version</key>
 	<integer>1</integer>
 	<key>pfm_last_modified</key>
-	<date>2024-08-23T02:21:00Z</date>
+	<date>2024-10-02T14:21:00Z</date>
 	<key>pfm_platforms</key>
 	<array>
 		<string>macOS</string>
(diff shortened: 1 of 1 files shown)


kevinmcox:
Nope, just need more time in the day.

Manifests/ManagedPreferencesApplications/com.twocanoes.xcreds.plist
```diff
@@ -1340,7 +1572,7 @@ changing “passwordID” to the correct element ID. If the value you typed into
 		</dict>
 		<dict>
 			<key>pfm_description</key>
-			<string>Name for Shares menu item. Default: "Shares".</string>
+			<string>Name for Shares menu item. Default: &quot;Shares&quot;.</string>
```
kevinmcox, line 1575: Is this an intentional change, it displays fine as is in both ProfileCreator and Imazing Profile Editor.

kevinmcox:
I notice the `pfm_app_min` key is not in use at all, which is what is used to specify the version of the app that started supporting a specific key.

Wouldn't we want to mark that all these new keys require Xcreds 5?

relgit:
Thanks for the update @twocanoes!

This could be a good opportunity to also give nicer titles to `idpHostName`, `idpHostNames`, and `passwordElementID`. Would you like to do that?

relgit (approved):

PR #713, opened 2024-10-15
Google drive AutomaticErrorReporting

moojomoore:
Adding the `AutomaticErrorReporting` key to the manifest. This is an undocumented key, but manages this setting in Google Drive. I wasn't sure if there was a good way to record that it's an undocumented key.

Screenshot 2024-10-15 at 8 50 19 AM


kevinmcox:
> I wasn't sure if there was a good way to record that it's an undocumented key.

It looks like we do it in this same manifest already with a `pfm_note`:
https://github.com/ProfileCreator/ProfileManifests/blob/master/Manifests/ManagedPreferencesApplications/com.google.drivefs.settings.plist#L367

kevinmcox (approved):

kevinmcox:
🎉  Thanks for the PR!

PR #720, opened 2024-11-05
Windows App

relgit:
This PR contains updates relating to the Microsoft Remote Desktop rename to Windows App.

For compatibility with profiles existing in the wild, the old Remote Desktop app key was not removed.


TVOps1:
The Windows App MSRD10 option is missing "Windows App" in the Application ID title line. This shows in Imazing as simply "Application ID"
MSRD10 Imazing

relgit:
Thanks for alerting us to that @TVOps1. It will be addressed separately.

kevinmcox (approved):

PR #721, opened 2024-11-05
Add SentinelOne Registration Token Manifest

drtaru:
This adds a payload for the new SentinelOne Registration Token domain to allow registering sensors on install without the need for a postinstall script.


kevinmcox:
Thanks for the PR @drtaru. Please remove the index files though, they are generated automatically after merge.

drtaru:
Apologies! Reverted that change!

kevinmcox (approved):

Issue #669, opened 2024-04-24
Add Okta Verify


homebysix:
### Please provide the following information:

**App Name**: Okta Verify

**App URL**: https://help.okta.com/en-us/content/topics/mobile/okta-verify-overview.htm

**App Profile Documentation URL**: https://help.okta.com/oie/en-us/content/topics/identity-engine/devices/managed-app-configs-macos.htm

**Payload / Domain**: `com.okta.mobile` and `com.okta.mobile.auth-service-extension`

Fgeronimo:
Since PR was merged, can we close this issue @kevinmcox @relgit? Thanks!

PR #737, opened 2024-12-26
Add Okta Verify

Fgeronimo:
Addresses https://github.com/ProfileManifests/ProfileManifests/issues/669

[Documentation](https://help.okta.com/oie/en-us/content/topics/identity-engine/devices/managed-app-configs-macos.htm)

I have a couple of questions:

- [This has been asked before](https://macadmins.slack.com/archives/CH9LT65U6/p1678477634031969), and I didn't address it yet. How should I proceed? Separate manifest files or something else?
- I'm confused by the documentation. Should [OktaVerify.Plugins](https://help.okta.com/oie/en-us/content/topics/identity-engine/devices/managed-app-configs-macos.htm#OktaVeri4) be included in this manifest? Should it be in the `com.crowdstrike.zta` preference domain?
- Updated in https://github.com/ProfileManifests/ProfileManifests/pull/737/commits/16d6d042ed09a10223a97970d618abc0e236a894  ~~Regarding the [Device Health Option](https://help.okta.com/oie/en-us/content/topics/identity-engine/devices/managed-app-configs-macos.htm#OktaVeri) preference key, I'm unsure if it will work as it expects values separated by semicolons. I’d appreciate it if anyone could test this to see if it applies correctly.~~

I appreciate any assistance you can provide. Thank you!


kevinmcox:
Thanks for the submission, Okta sure doesn't want to make this easy.

> I'm confused by the documentation. Should [OktaVerify.Plugins](https://help.okta.com/oie/en-us/content/topics/identity-engine/devices/managed-app-configs-macos.htm#OktaVeri4) be included in this manifest? Should it be in the com.crowdstrike.zta preference domain?

I think this should be in the Okta preference domain based on reading the documentation.

I want to hear what @relgit thinks about all the other questions.

relgit:
@kevinmcox is right, but I don't think that we'd be able to fully support the `OktaVerify.Plugins` key in the manifest because of its highly unusual data structure.

If I understand it correctly, besides the array of bundle IDs you also need to insert custom-named keys at the root of the dictionary. You could theoretically do that with a pair of `{{key}}` and `{{value}}` subkeys, but none of the profile editors support this at the root level.

Regarding the two separate domain strings, the best solution would be to have two manifests in tandem, with identical subkeys but different domains and meta.

In the discussion you linked, AP is probably referring to `pfm_value_copy`, which may provide the functionality that you are looking for but I don't know if it works across payloads, and it is only currently supported in ProfileCreator.

Cheers!

Fgeronimo:
Thank you both for the help and clarification @kevinmcox @relgit

> If I understand it correctly, besides the array of bundle IDs you also need to insert custom-named keys at the root of the dictionary. You could theoretically do that with a pair of {{key}} and {{value}} subkeys, but none of the profile editors support this at the root level.

Good catch! I just tested this by downloading Okta Verify from App Store and exporting a profile with the following settings in the profile installed locally on my system (EnrollmentOptions/OrgURL/ReportDiagnostics/LaunchOptions) and Okta Verify took the settings applied in the profile. I am not an Okta customer so I can't test DeviceHealth (it won't work in it's current state).

> Regarding the two separate domain strings, the best solution would be to have two manifests in tandem, with identical subkeys but different domains and meta.

Will work on and update once completed.

> In the discussion you linked, AP is probably referring to pfm_value_copy, which may provide the functionality that you are looking for but I don't know if it works across payloads, and it is only currently supported in ProfileCreator.

From what I understand from the docs, this does not span manifests

relgit (changes requested):
Looks like the manifests ended up not being identical in their subkeys. For example, `com.okta.mobile.plist `is missing `OktaVerify.UserPrincipalName` – why is that?

Also, they both lack a `PayloadOrganization` subkey.

Fgeronimo:
> Looks like the manifests ended up not being identical in their subkeys. For example, `com.okta.mobile.plist `is missing `OktaVerify.UserPrincipalName` – why is that?

@relgit I noticed that for `OktaVerify.UserPrincipalName`, I didn't see it referenced in the [docs](https://help.okta.com/oie/en-us/content/topics/identity-engine/devices/managed-app-configs-macos.htm). However, there is reference [here](https://help.okta.com/oie/en-us/content/topics/oda/macos-pw-sync/configure-macos-password-sync-policies.htm) as an optional key to be added in both manifests. When I tested this, `OktaVerify.UserPrincipalName` doesn't apply to Okta Verify, only Desktop Password Sync (from what I can see in testing). I also don't have strong feelings to leave UserPrincipleName out in `com.okta.mobile.plist` so if you want it in, I can do that.

> Also, they both lack a PayloadOrganization subkey.

Corrected in https://github.com/ProfileManifests/ProfileManifests/pull/737/commits/b9debda90dabbf0f4b8d881b7c0649808e442901

Fgeronimo:
Following up any other concerns here with this PR @relgit?

Manifests/ManagedPreferencesApplications/com.okta.mobile.auth-service-extension.plist
```diff
@@ -0,0 +1,336 @@
...
+			<key>pfm_type</key>
+			<string>string</string>
+		</dict>
+		<dict>
+			<key>pfm_description</key>
+			<string>Automatically populates the username in the Sign-In Widget</string>
+			<key>pfm_documentation_url</key>
+			<string>https://help.okta.com/oie/en-us/content/topics/oda/macos-pw-sync/configure-macos-password-sync-policies.htm</string>
+			<key>pfm_name</key>
+			<string>OktaVerify.UserPrincipalName</string>
```
relgit, line 316: Looks like this instance of `OktaVerify.UserPrincipalName` is a duplicate.

You can remove either this or the other one, depending on what order you think the field should appear in. If you remove this one, please remember to move its counterpart in the other manifest to match.

relgit (changes requested):

Manifests/ManagedPreferencesApplications/com.okta.mobile.auth-service-extension.plist
Fgeronimo: Apologies corrected in https://github.com/ProfileManifests/ProfileManifests/pull/737/commits/f025de28e318666a1fa7b9f7cd969d0768a9a360
OktaVerify.UserPrincipalName exists in other manifest. Appreciate the help here!

relgit (approved):

PR #740, opened 2024-12-30

doschupp:
Add new values for NSM v3.1.0 and newer


kevinmcox (changes requested):
Thanks for the contribution @doschupp.

Can you please run the file through plutil and validate the cleanup:
`plutil -convert xml1 de.fau.rrze.NetworkShareMounter.plist`

Please bump `pfm_version` since new keys were added.

Why is `useNewDefaultLocation` being removed? Normally we leave old keys for folks who may still be deploying old versions of the app. We just notate they they no longer apply with a `pfm_app_max` value.

kevinmcox:
Closing due to lack of response.

PR #752, opened 2025-01-23
Standardize and unify common payload keys

### LINKED ISSUE #751: Add PayloadOrganization subkey
**Name of App / Payload**: Various manifests (`org.videolan.vlc.plist`,`net.glencode.Particulars.plist`,`com.jigsaw24.ElevateSecurityLogs.plist`,etc)

**Preference key(s) to be added / updated**: `PayloadOrganization`

Upon correcting my mistake in this [PR](https://github.com/ProfileManifests/ProfileManifests/pull/737#issuecomment-2599052805), many manifests don't have PayloadOrganization subkey.

relgit:
This PR standardizes and unifies the [common payload keys](digidna/profilemani) across all active system and app manifests.

As part of this, the following work was done:
* Keys were added where missing. This is mainly about the `PayloadOrganization` key which was missing in many places, but others were missing too here and there.
* Keys were aligned with Apple's most recent documentation. This is largely about titles and descriptions but includes other information such as the range list of the `PayloadVersion` key.
* The order of keys was made consistent across all manifests in this PR.

@kevinmcox  – two points for you:

First, no changes were made to the _ManagedPreferencesDeveloper_ manifests. These are not used for actual configuration so I wanted to discuss what treatment they should get if any.

And second, the unified order of the keys means that the display name and description keys are now flipped in ProfileCreator. If that's ok we can proceed. Otherwise I could revert the order in the _Configuration.plist_ manifest as an exception.

The PR closes #751 (shout-out to @Fgeronimo for noticing the missing keys).

### DIFF
--- a/Manifests/ManagedPreferencesApple/com.apple.AdLib.plist
+++ b/Manifests/ManagedPreferencesApple/com.apple.AdLib.plist
@@ -20,7 +20,7 @@
 			<key>pfm_default</key>
 			<string>Configures Ad Tracking settings</string>
 			<key>pfm_description</key>
-			<string>Description of the payload.</string>
+			<string>The human-readable description of this payload. This description appears on the Detail screen.</string>
 			<key>pfm_description_reference</key>
 			<string>Optional. A human-readable description of this payload. This description is shown on the Detail screen.</string>
 			<key>pfm_name</key>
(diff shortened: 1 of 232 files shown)


kevinmcox (approved):
> And second, the unified order of the keys means that the display name and description keys are now flipped in ProfileCreator. If that's ok we can proceed. Otherwise I could revert the order in the Configuration.plist manifest as an exception.

Let's roll with it for now, we can always flip it back later if we decide to.

PR #758, opened 2025-02-07
Tailscale: Add Preferences + Min Versions

RVRX:
New Profiles for both `io.tailscale.ipn.macos.plist` &  `io.tailscale.ipn.macsys.plist`:
- **Add CLIIntegration to tailscale**
- **Add HideDHCP121Warnings to tailscale**
- **Add Hostname to tailscale**
- **Add VPNOnDemandSettings to tailscale**

Misc:
- **Add missing app_min values to tailscale**
- **fix incorrect minimum version in tailscale profile**
- **Bump modified+version**

### DIFF
--- a/Manifests/ManagedPreferencesApplications/io.tailscale.ipn.macos.plist
+++ b/Manifests/ManagedPreferencesApplications/io.tailscale.ipn.macos.plist
@@ -132,6 +132,8 @@ A profile can consist of payloads with different version numbers. For example, c
 			<string>string</string>
 		</dict>
 		<dict>
+            <key>pfm_app_min</key>
+			<string>1.52</string>
 			<key>pfm_description</key>
 			<string>Specifies the name of the organization managing Tailscale, for instance “XYZ Corp, Inc.”. The value will be displayed in the Tailscale client, so that users can easily reach your internal support resources.</string>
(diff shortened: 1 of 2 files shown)


RVRX:
Not all inclusive, there are still 1 or 2 preferences missing from what Tailscale fully supports. 

One I know is missing is [Suggest allowed forced exit nodes](https://tailscale.com/kb/1315/mdm-keys#suggest-allowed-forced-exit-nodes), which wants lists of strings or 'an empty set'. Wasn't immediately obvious how to implement that.

I may get to this and the other remaining profiles later.

relgit (approved):

relgit:
This looks all good to me @RVRX, thanks.

Regarding `AllowedSuggestedExitNodes` it's definitely non-conventional, but workable. Both iMazing Profile Editor and ProfileCreator allow adding empty arrays, so you can definitely add the key.

kevinmcox:
Hey @RVRX just checking to see if you want to add those other preferences you talked about, or if we should merge this PR as-is and add the other keys later.

Thanks!

RVRX:
@kevinmcox my bandwidth is quite low for now, will likely be a bit before I can get to it. Please go ahead and merge as is. Thanks

PR #759, opened 2025-02-10
Manifest update for XCreds 5.2

twocanoes:
Added keys for XCreds 5.2

### DIFF
--- a/Icons/index
+++ b/Icons/index
@@ -567,6 +567,20 @@
 			<key>path</key>
 			<string>Icons/ManagedPreferencesApplications/com.northpolesec.santa.png</string>
 		</dict>
+		<key>com.okta.mobile</key>
+		<dict>
+			<key>hash</key>
+			<string>4ddf4b38aa64491a463bed5ab2081aa3</string>
+			<key>path</key>
+			<string>Icons/ManagedPreferencesApplications/com.okta.mobile.png</string>
+		</dict>
+		<key>com.okta.mobile.auth-service-extension</key>
+		<dict>
+			<key>hash</key>
+			<string>4ddf4b38aa64491a463bed5ab2081aa3</string>
+			<key>path</key>
+			<string>Icons/ManagedPreferencesApplications/com.okta.mobile.auth-service-extension.png</string>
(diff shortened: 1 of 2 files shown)


Manifests/ManagedPreferencesApplications/com.twocanoes.xcreds.plist
```diff
@@ -1521,6 +1582,50 @@ changing “passwordID” to the correct element ID. If the value you typed into
...
+			<key>pfm_app_min</key>
+			<string>5.2</string>
+			<key>pfm_description</key>
+			<string>In some Active Directory environments, users do not use username@domain to login; they use a UPN suffix to make the username easier to use. This setting maps the UPN suffix to the correct AD domain name. For example, if you have an AD domain of foo.com but want users to sign in as user@bar.com, a UPN suffix of foo.com is created in AD and the user account is set to user@foo.com. This setting then would map foo.com to bar.com by setting the key upn to foo.com and the domain key to bar.com.</string>
+			<key>pfm_name</key>
+			<string>upnSuffixToDomainMappings</string>
+			<key>pfm_subkeys</key>
+			<array>
+				<dict>
+					<key>pfm_hidden</key>
```
relgit, line 1595: I'm not sure this provides any benefit here - does it make any difference for you? (I'm aware this is a ProfileCreator-only key).

Manifests/ManagedPreferencesApplications/com.twocanoes.xcreds.plist
```diff
@@ -24,7 +24,7 @@
 			<key>pfm_default</key>
 			<string>Configures XCreds configuration preferences</string>
 			<key>pfm_description</key>
-			<string>The human-readable description of this payload. This description appears on the Detail screen.</string>
```
relgit, line 27: Let's not make changes to the common payload keys (those stating with `Payload...`) because they are generic and except for their default values should be the same throughout all manifests.

relgit (changes requested):
Thanks for the updates @twocanoes, please see my two comments.

PR #772, opened 2025-03-24
Update webhook format in Jamf Setup Manager

fraserhess:


relgit:
Thanks for the update @fraserhess.

Have you tested without `pfm_hidden`? I believe it is not necessary inside of arrays.

fraserhess:
@relgit no, I started by copying and pasting from [the documentation](https://github.com/ProfileManifests/ProfileManifests/wiki/Array-of-Dictionaries) (also, [here](https://github.com/ProfileManifests/ProfileManifests/wiki/Preference-Templates#array-of-dictionaries)) where it says that portion is critical.

fraserhess:
I'm really unsure how to proceed here. I added `pfm_hidden` as it was described as critical in the documentation. While it _appears_ to work without it, the ProfileCreator code [still checks for it](https://github.com/ProfileCreator/ProfileCreator/blob/3ae18c86caec6f97bccc6dee7f67e1ec8bc52820/ProfileCreator/ProfileCreator/Profile%20Editor%20TableView%20CellViews/PayloadCellViewTableView.swift#L401) in some cases

relgit:
Thanks for linking to the docs and the code @fraserhess. I didn't want to make you wait but this was very fascinating and I needed to find an opportunity to dig into it.

So, looking at the code, I found that unfortunately the documentation is incorrect. The key is only important when dictionaries are nested under other dictionaries – not arrays. Kind of raises an eyebrow with how emphasized the importance of `pfm_hidden` is, but anyway, I'm glad that you pointed it out because now we have the option to fix it.

The next thing to do then is to remove the tag, and as soon as you do that I will merge the PR. Thanks!

PR #781, opened 2025-04-23
Update com.secondsonconsulting.baseline.plist

BigMacAdmin:
Adding new Baseline v2.3 features


relgit:
Thanks @BigMacAdmin. You'll notice that I changed the descriptions on the common payload keys so that they reflect Apple's latest, and you may want to sync that back to your original. Cheers!

BigMacAdmin:
Thanks @relgit for adding those back, I think you made a "mass update" during the time that I was developing and I had already copied the old values.

PR #784, opened 2025-04-29
App Auto Patch

TechTrekkie:
This is a profile manifest for the App Auto-Patch tool. With version 3.0 we moved to a configuration profile model and a manifest was created to assist Mac admins with configuring the tool for their environment. 

App Auto-Patch has been growing in popularity and the manifest has been quite useful so far so I felt it was time to get it added to the repo

You can find information about App Auto-Patch here:
[App Auto-Patch Github](https://github.com/App-Auto-Patch/App-Auto-Patch)
[App Auto-Patch Wiki](https://github.com/App-Auto-Patch/App-Auto-Patch/wiki)


Manifests/ManagedPreferencesApplications/xyz.techitout.appAutoPatch.plist
```diff
@@ -0,0 +1,1791 @@
...
+			<key>pfm_description</key>
+			<string>Forces the AAP dialogs to stay on top of all other windows</string>
+			<key>pfm_description_reference</key>
+			<string>https://github.com/App-Auto-Patch/App-Auto-Patch/wiki/Configure-Settings#dialogontop</string>
+			<key>pfm_range_list</key>
+			<array>
+				<string>TRUE</string>
+				<string>FALSE</string>
+			</array>
+			<key>pfm_range_list_titles</key>
```
relgit, line 426: If the titles match the actual values then there's no need for range titles (`pfm_range_list_titles`). I would however recommend to keep it and change the title to something more suitable for display. For example, here you can change `TRUE` and `FALSE` to Force and Do Not Force.

relgit (changes requested):
Thanks for the PR @TechTrekkie. Always nice to see a well made manifest.

I have one specific comment about the use of ranged string preferences which you will see in the code.

After you make changes, please run the file through `plutil -convert xml1 [file name]` to standardize it.

One general recommendation that I have is to consider using the plist boolean type in the future for true/false values. This way they will be displayed as checkboxes in the profile editors, and users would only need one click to set them up.

Cheers!

TechTrekkie:
Hi @relgit ! Thanks for the input. I've made some updates to several of the range titles (`pfm_range_list_titles`) to put some more suitable display titles. I also ran it through standardization. Let me know if that all works or if any additional changes need to be made. 
For future versions, I'll work on updating the `TRUE` and `FALSE` options to boolean after I review and test with my script. 

Thank you!

TechTrekkie (commented):
I updated the file in my repo with updated range titles

relgit:
Thank you @TechTrekkie!

PR #790, opened 2025-05-20
updated manifest for XCreds 5.4

twocanoes:
Note: the app manifest number skips 12 (from 11 to 13) because we did not do a pull request for xcreds 5.3. We would like to the version since version 12 exists other places.


relgit (changes requested):
Thank you for the update @twocanoes. Here are a handful of changes I'd like to ask.

Please replace the use of `float` in `pfm_type` with `real`.

I've noticed that neither `HomeMountEnabled` nor `HomeAppendDomain` have default values. Would you like to add those? We can also discuss adding conditions to exclude these properties if you like since I see that they depend on others.

Lastly, please run the manifest through `plutil -convert xml1` to standardize it.

It's always nice when developers contribute their manifests and we appreciate you continuing to do so.

PR #791, opened 2025-05-22
feat: Add dev.firezone.firezone.plist 🎉

jamilbk:


jamilbk:
Moved to draft pending the release of our 1.4.15 client which supports the configuration keys here. We expect that to land in a day or so.

relgit (changes requested):
Thank you for the submission @jamilbk. I have just a few changes to ask for.

Please add default values to the following boolean properties, if there are such:
`disableUpdateCheck`
`startOnLogin`
`connectOnStart`
`hideAdminPortalMenuItem`
`internetResourceEnabled`

Please remove the `pfm_description_reference` key when it is identical to `pfm_description`.

And lastly, please pass the manifest through the command `plutil -convert xml1` to standardize it.

Cheers 🙂

relgit (approved):
Thanks for making the changes.

jamilbk:
@relgit Thanks for the approval! Before we release the client and I mark this ready, I had a question around the default for the boolean fields.

By default, we want these to remain `Unset`. If the admin chooses a particular option, that will override the user-selected equivalent in the Firezone Client.

Is assigning a default value to these still the right approach, or should we remove the `pfm_default`s for these? Based Is `boolean` the correct approach here from a UX perspective? I noticed some apps use a `string` field here with allowed values which improves things in the iMazing Profile Editor, but adds more complexity for admins generating the `mobileconfig` using other tools or from scratch.

jamilbk:
@relgit Nevermind on my previous questions - I simply updated the descriptions to better explain what setting each field does.

Our 1.4.15 clients have shipped, so this is now ready to go.

jamilbk:
Marked as draft while I get our icon added.

PR #795, opened 2025-05-29
Fix typo in Okta Verify SSO Extension (#794)

### LINKED ISSUE #794: Okta Verify SSO Extension label typo
**Name of App / Payload**: Okta Verify (SSO Extension)

**Preference key(s) needing fixing**: `OktaVerify.UserPrincipalName`

**Description of problem(s) / error(s) with listed key(s)**: The label for this property uses the erroneous text “User Principle Name”.

**Description of expected behavior**: This text should be changed to say “User Principal Name”, both because that is the correct spelling and because it matches the name of the underlying key.

**How are you testing / deploying your profile?**: N/A, noticed this in ProfileCreator UI

wjk:
Closes #794. Date has been updated as documented in the wiki. Enjoy!


wjk:
Unfortunately, I forgot to account for time zones when I updated the timestamp. I based the time on my Mac’s clock, which is set to US Eastern Time. However, given that the date changed as well, I do not think this will be that big a problem.

relgit:
Thanks for the PR @wjk!

Yes, there aren't any side effects to the timestamp being off by a bit as long as it's more recent then the previous one.

PR #799, opened 2025-06-17
Update com.apple.TCC.configuration-profile-policy to Authorization Strings

tully-systima:
## **Name of App / Payload**:
`com.apple.TCC.configuration-profile-policy`

## **Preference key(s) to be added / updated**: 
- Updated all `Allowed` booleans to `Authorization` strings, and added documentation links and description of the expected behavior.
- The legacy boolean standard provides fewer options compared to the newer Authorization string standard, all pfm keys in `com.apple.TCC.configuration-profile-policy` have been updated to strings and include `Allow`, `Deny`, or `AllowStandardUserToSetSystemService`, as applicable to each [Property identity](https://developer.apple.com/documentation/devicemanagement/privacypreferencespolicycontrol/services-data.dictionary) requirement.
- Older boolean values do not allow selection of additional configuration options available via the newer String standard, and can sometimes cause errors when uploading to some MDM providers, error stating "Authorization is missing"
- As observed in testing, the `AllowStandardUserToSetSystemService` configuration option is valid and works as expected, including in Tahoe beta, therefore has been added to the available options for the `ListenEvent` and `ScreenCapture` payloads.
Apple documentation for this setting is currently inconsistent (see below for more information)
- Screenshot 2025-06-17 at 6 11 34 PM

- `Camera` and `Microphone` access only include `Deny` as per documentation

## **Applicable Preference Documentation URL(s)**:
NOTE: Apple documentation is currently inconsistent for `ListenEvent` and `ScreenCapture`:
- [PrivacyPreferencesPolicyControl.Services](https://developer.apple.com/documentation/devicemanagement/privacypreferencespolicycontrol/services-data.dictionary) currently incorrectly states:
> Access to these events can’t be given in a profile; it can only be denied.
- However, [PrivacyPreferencesPolicyControl.Services.Identity](https://developer.apple.com/documentation/devicemanagement/privacypreferencespolicycontrol/services-data.dictionary/identity) states:
> AllowStandardUserToSetSystemService: Allows a standard (non-admin) user to configure the permissions for the specified app in the Privacy preferences for services that otherwise require admin authorization; only valid for the ListenEvent and ScreenCapture services
- As above, `AllowStandardUserToSetSystemService` configuration option is valid and works as expected


relgit:
Thank you for the PR @tully-systima. Could you please run the manifest file through `plutil -convert xml1` to standardize it please?

tully-systima:
@relgit  Done!

relgit:
@kevinmcox there's one side effect to this work for ProfileCreator users that I think we should be aware of. Removing the old `Allowed` keys and replacing them with the new `Authorization` ones will diminish support for PPPC profiles created up until the change since their `Allowed` values won't be shown anymore.

I wish we could keep both the old and the new, but it looks like ProfileCreator will then configure both, and in this domain Apple specifically [forbids](https://developer.apple.com/documentation/devicemanagement/privacypreferencespolicycontrol/services-data.dictionary/identity) that.

Of course, there are downsides to not merging the work. One of these is users loading profiles from other sources like their MDM, which may already be using the new keys, and not seeing these keys right now. Other reasons are laid out by @tully-systima above.

I think the downsides of not merging are greater than those of merging. What do you think?

tully-systima:
Note: @relgit @kevinmcox I have found another inconsistency that should be considered:

I have just noticed that [in the master branch](https://github.com/ProfileManifests/ProfileManifests/blob/257e95c3ceaa7c5a25198c1b1c61bd9ae14f9d2a/Manifests/ManifestsApple/com.apple.TCC.configuration-profile-policy.plist#L2909), the TCC PFM is already a mixed bag of Allowed bools and Authorization strings. Whilst most items are on `Allowed` bools, both `BluetoothAlways` and `SystemPolicyAppData` are currently set to use `Authorization` strings, and `SystemPolicyAppData` also includes `AllowStandardUserToSetSystemService` as an option, which I believe is a mistake, as that is not an available option for SystemPolicyAppData. [Documentation](https://developer.apple.com/documentation/devicemanagement/privacypreferencespolicycontrol/services-data.dictionary/identity) currently states:
> AllowStandardUserToSetSystemService: ...  [is] only valid for the **ListenEvent** and **ScreenCapture** services

... whilst the `ScreenCapture` and `ListenEvent` items are still on `Allowed` bools , and therefore cannot use `AllowStandardUserToSetSystemService`

This also means that profiles generated with this PFM, if either `BluetoothAlways` or `SystemPolicyAppData` is included in the payload, will then create a broken profile, as Apple states you cannot use both Approved bools and Auth strings simultaneously.

At the end of the day, when the rubber meets the road, the desired end result should always be "what makes our users lives better", and the reality is that if we stay on the old `Allowed` bools without `AllowStandardUserToSetSystemService`, then a user trying to do a presentation and share their screen over zoom / slack etc, needs to then call IT in the middle of the meeting and get them to input an admin password to allow it... which has been a source of a lot of frustration from my clients.

kevinmcox:
Sorry for the delay, it'll be a couple days until I'm able to take a look and digest this.

tully-systima:
@kevinmcox touching base, any thoughts on this?

relgit:
@tully-systima I'm going to merge your work into a side-branch where I'm working on preparing the repository for the version 27.0 generation. It should eventually get merged along. Thanks!

PR #803, opened 2025-06-30
Relocate ServerPort key as a subkey to SyslogOptions

lhaeger:
ServerPort did not apply because it been in the wrong position in the tree. See https://github.com/SAP/macOS-enterprise-privileges/wiki/Managing-Privileges#SyslogOptions for details


relgit:
@lhaeger have you seen my comment?

PR #804, opened 2025-07-03
Don't modify Localizable.strings in situ

wjk:
The `plistlib` module can read binary plists. While I no longer have a Python 2 installation to test the script with, I believe that this feature is present in Python 2.

Doing this also breaks the Google Chrome app's code signature.

### DIFF
--- a/Resources/ManifestsGenerated/Google Chrome/chromeManifest.py
+++ b/Resources/ManifestsGenerated/Google Chrome/chromeManifest.py
@@ -137,6 +137,6 @@ newmanifest["pfm_platforms"] = ["macOS"]
 # Add pfm_interaction
 newmanifest["pfm_interaction"] = "combined"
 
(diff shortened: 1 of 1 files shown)


relgit:
It's quite a find that the script modifies the app bundle in place – we definitely shouldn't have something like that.

Would you mind pushing your Python 3 porting attempt so I understand the issue better? I don't see long descriptions as a problem here.

relgit:
Alright I'm seeing the issue now. I've made a change to the script to not stop on such keys, but that does indeed mean that their text may turn up a bit wonky. That's alright though since such keys are in the minority and we'll be able to catch them and manually adjust when reviewing the diff.

I also made another commit changing the output of the script to its own folder rather then to the desktop. Somehow I wasn't really comfortable with it writing/overwriting freely on the desktop.

Let me know if you have anything to add @wjk. Otherwise I think it's ready for merging.

relgit (approved):

PR #832, opened 2025-10-20
Added manifest for Cursor

targendaz2:


kevinmcox (changes requested):
@targendaz2 this looks like a duplicate of #833.

Can you split these PRs up so one of them contains VS Code and the other contains Cursor please.

Issue #817, opened 2025-08-22
Chrome manifest settings missing


patgmac:
### Please provide the following information:

**Name of App / Payload**: *ex. Firefox OR org.mozilla.firefox*

Google Chrome

**Preference key(s) to be added / updated**: *ex. `AppUpdateURL`, `Bookmarks`, etc.*

DefaultClipboardSetting
SafeBrowsingForTrustedSourcesEnabled
RemoteAccessHostAllowUiAccessForRemoteAssistance
ExtensionUnpublishedAvailability
StrictMimetypeCheckForWorkerScriptsEnabled
RemoteDebuggingAllowed
RendererAppContainerEnabled
AllowWebAuthnWithBrokenTlsCerts
DefaultInsecureContentSetting
PdfLocalFileAccessAllowedForDomains
DomainReliabilityAllowed
EncryptedClientHelloEnabled
EnterpriseProfileCreationKeepBrowsingData
FileOrDirectoryPickerWithoutGestureAllowedForOrigins
GoogleSearchSidePanelEnabled
HttpsUpgradesEnabled

**Applicable Preference Documentation URL(s)**:

https://chromeenterprise.google/policies/

relgit:
Completed in #859.

PR #859, opened 2025-12-25
Update Google Chrome Manifest

Fgeronimo:
Related: https://github.com/ProfileManifests/ProfileManifests/issues/817

Tested in ProfileCreator

Note: For such a large change, I didn't sync changes to Brave for this PR. Let me know if I should sync to Brave in this PR or make a separate PR. Thank you!


Fgeronimo:
@relgit Can you review and let me know if changes need to be made. I'm missing something but when I submit this PR, I couldn't assign reviewers

relgit:
Thanks for the PR @Fgeronimo.

If you want to go ahead and sync to Brave that would be appreciated. We have a [script](https://github.com/ProfileManifests/ProfileManifests/blob/master/Resources/ChromeToBrave/ChromeToBrave.py) that makes it very easy by the way.

Ah, one more thing – having the separation into individual commits helped me in reviewing this so thanks for that.

Cheers!

PR #875, opened 2026-03-22
Update com.okta.mobile: add OktaVerify.Plugins key and update DeviceHealthOptions

anzzyspeaksgit:
## Summary
Adds `OktaVerify.Plugins` and updates `OktaVerify.DeviceHealthOptions` to include missing documented values.

## Problem
Closes #869

## Solution
- `OktaVerify.Plugins` (Array type) is documented by Okta but was missing from the manifest. It enables EDR client integration (e.g., CrowdStrike ZTA).
- `OktaVerify.DeviceHealthOptions` note/placeholder was missing some documented values: `HideOSUpdate` and `HidePassword`.
- Regenerated indexes using `updateIndexes.py`.

## Testing
- Verified plist structure
- Ran index updater without errors

---
*This PR was created with AI assistance.*

### DIFF
--- a/Icons/index
+++ b/Icons/index
@@ -1580,6 +1580,6 @@
 		</dict>
 	</dict>
 	<key>date</key>
-	<date>2026-03-09T10:15:46Z</date>
+	<date>2026-03-22T18:02:03Z</date>
 </dict>
 </plist>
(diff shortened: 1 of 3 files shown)


relgit (changes requested):
Thank you for the PR @anzzyspeaksgit.

I'm sorry but if I understand the documentation correctly, we cannot currently fully support `OktaVerify.Plugins`. The reason being that although it's a simple list of bundle IDs as you are showing in your PR, the property itself requires a **root** custom subkey to be configured alongside for each item.

That could be defined using the `{{key}}` and `{{value}}` convention, but realistically none of the profile editors support this at the root level right now. If we accept just the Plugins subkey, it may cause confusion and frustration for users trying to configure it, so I'd rather avoid it altogether.

As per the device health options, we can definitely keep those.

Apart for `OktaVerify.Plugins`, please also remove changes to the indexes as they are automatically generated on merge.

PR #882, opened 2026-04-14
Update com.dare.zappl.preferences.plist | New Keys, Revised Defaults, Updated Documentation Links

dellis-dare:
Updated com.dare.zappl.preferences.plist:

- Change documentation URLs
- Rename UI sections to "Scheduled/Forced Update Prompt Options" and update segment keys.
- Add new settings (Enable Compliance Reporting, Background Update Stagger Window, Automatically Install AppBar) and patchDayWeeks with week selection subkeys.
- Clarify and reword multiple descriptions and app catalog links, standardise "app/applications" phrasing, and adjust activation/public key titles.
- Increase prompt refresh defaults/minimums in scheduled and forced update prompts.
-  Bump pfm_version to 2.


Manifests/ManagedPreferencesApplications/com.dare.zappl.preferences.plist
```diff
@@ -131,14 +131,14 @@ A profile can consist of payloads with different version numbers. For example, c
 			<key>pfm_range_list_titles</key>
 			<array>
 				<string>General Options</string>
-				<string>Scheduled Update Options</string>
-				<string>Forced Update Options</string>
+				<string>Scheduled Update Prompt Options</string>
```
relgit: What's the reason for adding 'Prompt' to the segments? We usually try to keep them as short as possible. Come to think of it, the 'Options' suffix on all segments also seems a bit redundant to me now.

Manifests/ManagedPreferencesApplications/com.dare.zappl.preferences.plist
```diff
@@ -1301,21 +1416,21 @@ A profile can consist of payloads with different version numbers. For example, c
 				</dict>
 				<dict>
 					<key>pfm_default</key>
-					<integer>20</integer>
+					<integer>60</integer>
```
relgit, line 1569: Did the prompt refresh defaults minimums change in a recent new version? If so, it's worth mentioning in the description or a note for admins that load previously-made profiles which will now get a validation error.

relgit (changes requested):
Thanks for updating the manifest @dellis-dare,

Please see the two comments I made in the code. I also have two other general points.

First, please add `pfm_app_min` to all new properties if they were added in a more recent version of the app.

And the second is that I noticed that `pfm_app_url` is set to a download URL. Could you change that to the marketing website URL? That's what we have for all other apps.

Cheers and thank you 😀

Manifests/ManagedPreferencesApplications/com.dare.zappl.preferences.plist
dellis-dare: Totally fair question! I did this to match our segments in our Custom Schema we use with Jamf, but you've made me question that now too... 😅
I'll update segment strings to `General`, `Scheduled Updates` and `Forced Updates`.

Manifests/ManagedPreferencesApplications/com.dare.zappl.preferences.plist
dellis-dare: I'm verifying with our developer but I will add to the existing `pfm_description` with a note on the default and minimum values having changed, and when they did so.

dellis-dare:
Thanks for taking the time to review!

- Added `pfm_app_min` to the relevant keys
- Altered `pfm_app_url` to the product homepage on our website
- Shortened Segment Titles as correctly suggested
- Clarified new default and minimum values, and when they were introduced, in the note of the `refreshInterval` key.

I also took the opportunity to add `pfm_documentation_url` keys to relevant keys, as I hadn't initially added those. Hopefully that's ok in this PR ☺️

Finally, please ignore https://github.com/ProfileManifests/ProfileManifests/pull/882/commits/799cddc8fe2cc17378a281b70fe5d66a9ef305c2 - I'm still finalising this new manifest for Zappl's companion app, but will raise it in a new PR when ready.

Thanks again!

relgit (approved):

Issue #862, opened 2026-01-05
Add 1Password 8 new keys


Fgeronimo:
### Please provide the following information:

**Name of App / Payload**: 1Password 8

**Preference key(s) to be added / updated**: authentication.defaultDomain / authentication.enforceDomain (will check for others)

**Applicable Preference Documentation URL(s)**: https://support.1password.com/mobile-device-management/?mac#authentication-opm8

relgit:
Completed in #885.

PR #885, opened 2026-04-22
Update 1Password8 Manifest: authentication.defaultDomain / authentication.enforceDomain

Fgeronimo:
Related: https://github.com/ProfileManifests/ProfileManifests/issues/862

Tested in ProfileCreator 0.3.7

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.1password.1password.plist
+++ b/Manifests/ManagedPreferencesApplications/com.1password.1password.plist
@@ -192,7 +192,6 @@ During a profile replacement, the system updates payloads with the same 'Payload
 			<string>boolean</string>
 		</dict>
 		<dict>
-			
 			<key>pfm_description</key>
 			<string>Set the default sign-in address for the 1Password app. Users will be prompted to sign in to this account by default.</string>
 			<key>pfm_documentation_url</key>
--- a/Manifests/ManagedPreferencesApplications/com.1password.1password.plist
+++ b/Manifests/ManagedPreferencesApplications/com.1password.1password.plist
@@ -127,6 +127,7 @@ During a profile replacement, the system updates payloads with the same 'Payload
 			<string>PFC_SegmentedControl_0</string>
 			<key>pfm_range_list_titles</key>
 			<array>
+				<string>Authentication</string>
 				<string>General</string>
 				<string>Privacy</string>
 				<string>Security</string>
@@ -136,6 +137,11 @@ During a profile replacement, the system updates payloads with the same 'Payload
 			<string>always</string>
 			<key>pfm_segments</key>
 			<dict>
+				<key>Authentication</key>
+				<array>
+					<string>authentication.defaultDomain</string>
+					<string>authentication.enforceDomain</string>
+				</array>
 				<key>General</key>
 				<array>
 					<string>app.startAtLogin</string>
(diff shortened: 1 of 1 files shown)


Manifests/ManagedPreferencesApplications/com.1password.1password.plist
```diff
@@ -185,6 +191,40 @@ During a profile replacement, the system updates payloads with the same 'Payload
 			<key>pfm_type</key>
 			<string>boolean</string>
 		</dict>
+		<dict>
+			<key>pfm_default</key>
+			<string>domain.1password.com</string>
```
relgit: This should use `pfm_value_placeholder` instead of `pfm_default`.

Manifests/ManagedPreferencesApplications/com.1password.1password.plist
```diff
@@ -185,6 +191,40 @@ During a profile replacement, the system updates payloads with the same 'Payload
...
+		</dict>
+		<dict>
+			<key>pfm_default</key>
+			<false/>
+			<key>pfm_description</key>
+			<string>When enabled, users cannot change the sign-in address from the default value set in authentication.defaultDomain.</string>
+			<key>pfm_documentation_url</key>
+			<string>https://support.1password.com/mobile-device-management/?mac#authentication-opm8</string>
+			<key>pfm_name</key>
+			<string>authentication.enforceDomain</string>
```
relgit, line 232: Looks like we need to add a `pfm_exclude` here (see [examples](https://github.com/ProfileManifests/ProfileManifests/wiki/Manifest-Format#example-conditions--exclusions)) because the previous property is required for this one.

If you're not familiar with setting this up I can help.

Manifests/ManagedPreferencesApplications/com.1password.1password.plist
```diff
@@ -185,6 +191,40 @@ During a profile replacement, the system updates payloads with the same 'Payload
...
+			<key>pfm_note</key>
+			<string>Use the format: domain.1password.com. Do not include https:// or trailing slashes.</string>
+			<key>pfm_title</key>
+			<string>Default Sign-in Address</string>
+			<key>pfm_type</key>
+			<string>string</string>
+		</dict>
+		<dict>
+			<key>pfm_default</key>
+			<false/>
```
relgit: What's the source for this? I don't see a mention of a default value in the docs. Testing can definitely be a source though.

relgit (changes requested):
Thanks for the PR @Fgeronimo. Please have a look at my comments.

Manifests/ManagedPreferencesApplications/com.1password.1password.plist
Fgeronimo: Corrected in https://github.com/ProfileManifests/ProfileManifests/pull/885/commits/72734f76cc3bf27b3530bc53747e34436bae5843

Manifests/ManagedPreferencesApplications/com.1password.1password.plist
Fgeronimo: @relgit Help would be appreciated here. I tried a few different `pfm_target_conditions` keys but can't get it to work.

Manifests/ManagedPreferencesApplications/com.1password.1password.plist
Fgeronimo: From testing left it on accident removed in https://github.com/ProfileManifests/ProfileManifests/pull/885/commits/f9ba46d35cb0d54503032cac7c020a0c971d567b

Manifests/ManagedPreferencesApplications/com.1password.1password.plist
relgit: This one's not on you actually. The reason it's not working is because the target name contains a period, which is a KVC path separator.

Essentially, when a profile editor looks for the target, it first looks for an `authentication` subkey and then for a `defaultDomain` child. Since this combination does not exist, setting up the condition fails.

I still added it though, for posterity and since support for this could be added in the future.

PR #890, opened 2026-05-01
Update Claude Desktop Manifest: isDesktopExtensionSignatureRequired / disabledBuiltinTools / managedMcpServers

homebysix:
Adds three MDM keys to the Claude Desktop manifest, documented by Anthropic at
https://support.claude.com/en/articles/14680753-extend-claude-cowork-with-third-party-platforms
but not yet present in this manifest.

### Notes

- `headers` and `toolPolicy` (inside `managedMcpServers`) are modeled as open `dictionary` types since both take arbitrary keys. Allowed `toolPolicy` values (`allow`, `ask`, `blocked`) are noted in the description.
- `disabledBuiltinTools` is left as a free-form string array (with known tool names listed in the description) rather than constrained to an enum, so it doesn't go stale when Anthropic adds new tools.
- No `pfm_app_min` set on the new keys — Anthropic's docs don't state when they shipped.
- `pfm_version` bumped 2 → 3, `pfm_last_modified` updated.

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.anthropic.claudefordesktop.plist
+++ b/Manifests/ManagedPreferencesApplications/com.anthropic.claudefordesktop.plist
@@ -13,7 +13,7 @@
 	<key>pfm_format_version</key>
 	<integer>1</integer>
 	<key>pfm_last_modified</key>
-	<date>2026-02-24T10:11:42Z</date>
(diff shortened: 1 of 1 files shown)


relgit:
Thanks for the PR @homebysix. I took the liberty of changing the dictionary subkey names to fit convention, to add range list titles, and to flesh out the custom dictionaries.

PR #891, opened 2026-05-03
Fix "exlusive" typo in pfm_interaction (3 files)

homebysix:
Three manifests had pfm_interaction set to "exlusive" (missing the c) instead of "exclusive", which silently falls back to the default behavior — the opposite of what the authors intended. Corrected the value in com.apple.controlcenter, com.apple.preferences.sharing.SharingPrefsExtension, and com.apple.mcxprinting, and bumped pfm_version + pfm_last_modified for each.

Confirmed the typo is not present in Apple's canonical apple/device-management spec, so this is purely a ProfileManifests-side error.


relgit:
Thanks, @homebysix. Quick note that I've reverted the version change – we normally only bump that when subkeys are added or removed.

PR #892, opened 2026-05-03
Fix Parallels Desktop filename typo

homebysix:
The manifest for `com.parallels.desktop.managedprefs` was filed as `com.parallels.dektop.managedprefs.plist` (missing the `s` in "desktop"), even though the `pfm_domain` inside the file is spelled correctly. Renamed the file to match its domain so it sorts and is discoverable as expected, and bumped pfm_version + pfm_last_modified.


relgit:
Thank you @homebysix

PR #893, opened 2026-05-03
Fix Baseline Retries integer defaults

homebysix:
The five nested `Retries` keys in com.secondsonconsulting.baseline are declared as `pfm_type` integer but had `<false/>` as their `pfm_default`, which is a type mismatch. Changed them to `<integer>0</integer>` to match the sibling top-level `DefaultInstallomatorRetries`/`DefaultScriptRetries`/`DefaultPackageRetries` already defined in the same manifest, and bumped pfm_version + pfm_last_modified.


relgit (approved):

relgit:
Thanks you @homebysix

PR #894, opened 2026-05-03
Fix ProxyCaptiveLoginAllowed integer default

homebysix:
The `ProxyCaptiveLoginAllowed` subkey under `com.apple.SystemConfiguration` Proxies is declared as `pfm_type` integer (matching Apple's canonical device-management spec, which lists `type: <integer>` and the description "If 1, allows client to log into captive portal network"), but its `pfm_default` was `<false/>` — a type mismatch. Changed to `<integer>0</integer>`, which matches the 15 sibling integer-typed keys in the same manifest, and bumped pfm_version + pfm_last_modified.


relgit:
Thanks @homebysix

PR #897, opened 2026-05-05
Update to support Munki 7.1 features

homebysix:
I've added the preference keys newly supported by [Munki 7.1](https://github.com/munki/munki/wiki/Munki-7.1-Introduction).

### DIFF
--- a/Manifests/ManagedPreferencesApplications/ManagedInstalls.plist
+++ b/Manifests/ManagedPreferencesApplications/ManagedInstalls.plist
@@ -13,7 +13,7 @@
 	<key>pfm_format_version</key>
 	<integer>1</integer>
 	<key>pfm_last_modified</key>
-	<date>2024-08-29T07:55:25Z</date>
+	<date>2026-05-05T00:00:00Z</date>
 	<key>pfm_platforms</key>
 	<array>
 		<string>macOS</string>
@@ -184,6 +184,9 @@ During a profile replacement, the system updates payloads with the same 'Payload
 					<string>AggressiveUpdateNotificationDays</string>
 					<string>DaysBetweenNotifications</string>
 					<string>InstallRequiresLogout</string>
+					<string>MSCOfferToQuitBlockingApps</string>
+					<string>MSCOfferToForceQuitBlockingApps</string>
+					<string>MSCOfferToUpdateOthers</string>
 					<string>ShowRemovalDetail</string>
 					<string>ShowOptionalInstallsForHigherOSVersions</string>
 					<string>SuppressUserNotification</string>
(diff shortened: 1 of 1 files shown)


Manifests/ManagedPreferencesApplications/ManagedInstalls.plist
```diff
@@ -536,6 +649,76 @@ During a profile replacement, the system updates payloads with the same 'Payload
...
+		<dict>
+			<key>pfm_app_min</key>
+			<string>7.1</string>
+			<key>pfm_conditionals</key>
+			<array>
+				<dict>
+					<key>pfm_target_conditions</key>
+					<array>
+						<dict>
+							<key>pfm_range_min</key>
```
relgit: There's no `pfm_range_min` key for target conditions. See [here](https://github.com/ProfileManifests/ProfileManifests/wiki/Manifest-Format#conditions).

Manifests/ManagedPreferencesApplications/ManagedInstalls.plist
```diff
@@ -536,6 +649,76 @@ During a profile replacement, the system updates payloads with the same 'Payload
...
+			<string>How many seconds to wait between download retries. Valid values are 1-30.</string>
+			<key>pfm_documentation_url</key>
+			<string>https://github.com/munki/munki/wiki/Munki-7.1-download-retries</string>
+			<key>pfm_exclude</key>
+			<array>
+				<dict>
+					<key>pfm_target_conditions</key>
+					<array>
+						<dict>
+							<key>pfm_present</key>
```
relgit, line 659: This should probably be changed to exclude when `DownloadRetries` is not true (`pfm_n_range_list`).

relgit (changes requested):
Thanks for the PR @homebysix. Please have a look at my comments.

Manifests/ManagedPreferencesApplications/ManagedInstalls.plist
homebysix: Done in 045b4bc4.

Manifests/ManagedPreferencesApplications/ManagedInstalls.plist
homebysix: Done in 045b4bc4.

relgit:
@homebysix I was wrong about the target condition in `DownloadRetrySleepSeconds` because `DownloadRetries` is actually an integer. So, it's now reverted.

I also removed a handful of requirement conditionals that I don't think are warranted, and fixed two target conditions so that they will work even when the target has no value.

PR #900, opened 2026-05-08
Zappl | Adding conditional pfm keys

dellis-dare:
Hello, me again ☺️

We've received some community feedback that highlighted a need for us to show/hide certain preferences that are reliant on the values of other keys.
This PR includes the addition of `pfm_exclude` and `pfm_target_conditions` keys where relevant, as well as `pfm_note` keys to help describe the conditions that need to be met.

As always, I appreciate the time and support provided!

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.dare.zappl.preferences.plist
+++ b/Manifests/ManagedPreferencesApplications/com.dare.zappl.preferences.plist
@@ -13,7 +13,7 @@
 	<key>pfm_format_version</key>
 	<integer>1</integer>
 	<key>pfm_last_modified</key>
-	<date>2026-04-22T00:00:00Z</date>
+	<date>2026-04-30T00:00:00Z</date>
 	<key>pfm_platforms</key>
 	<array>
 		<string>macOS</string>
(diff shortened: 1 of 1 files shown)


relgit:
Thanks for the PR @dellis-dare.

A quick question to understand if the conditions are correct: would `frequencyOption` (for example) work if `runType` is not specified?

If not, you should use `pfm_n_range_list` in the target condition (with the value of `specifyFrequency`) instead. This is generally the better way to set conditions that describe the desired value, and not the other possible values.

dellis-dare:
> A quick question to understand if the conditions are correct: would `frequencyOption` (for example) work if `runType` is not specified?

Thanks for checking! With that example, `runType` has a default value of `specifyFrequency`, of which `frequencyOption` is a child. So it should work if no value is specified, or if `specifyFrequency` is specified, which matches the current behaviour.

relgit:
Ok, in that case `frequencyOption` can keep its target conditions, but other properties, those that should hide when the `runType` value is unset or set to its default should use `pfm_n_range_list` logic.

For example, `patchDay` should have the following `pfm_exclude` block:
```
<key>pfm_exclude</key>
<array>
	<dict>
		<array>
			<dict>
				<key>pfm_n_range_list</key>
				<array>
					<string>specifyPatchDay</string>
				</array>
				<key>pfm_target</key>
				<string>ScheduledUpdates.runFrequency.runType</string>
			</dict>
		</array>
	</dict>
</array>
```

This will reduce the clutter when adding a new payload by initially displaying only the properties that go with `runType`'s default value.

dellis-dare:
Thank you for the feedback and considerations, @relgit! Much appreciated.

Those changes should be in place now, there was also a duplicate `pfm_target_conditions` array in `mondayStartTime` that has been removed.

relgit (approved):
Thanks for making the changes @dellis-dare. Cheers!

PR #903, opened 2026-05-18
Add forceLoginOrgUUID key to com.anthropic.claudefordesktop

rmalary:
Adds the `forceLoginOrgUUID` preference key to the Claude Desktop manifest. When set via MDM, this forces users to sign in with a specific Anthropic organization, enforcing enterprise authentication.

Validated on macOS with Jamf Pro using the Application & Custom Settings > Upload method. The preference lands correctly in `/Library/Managed Preferences/com.anthropic.claudefordesktop` and is read by Claude Desktop on next launch.

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.anthropic.claudefordesktop.plist
+++ b/Manifests/ManagedPreferencesApplications/com.anthropic.claudefordesktop.plist
@@ -133,8 +133,6 @@
 		<dict>
 			<key>pfm_description</key>
 			<string>Forces users to sign in with the specified Anthropic organization. When set, Claude Desktop requires the user to authenticate with the matching organization UUID.</string>
-			<key>pfm_documentation_url</key>
-			<string>https://code.claude.com/docs/en/settings</string>
 			<key>pfm_format</key>
 			<string>^[0-9A-Za-z]{8}-[0-9A-Za-z]{4}-[0-9A-Za-z]{4}-[0-9A-Za-z]{4}-[0-9A-Za-z]{12}$</string>
 			<key>pfm_name</key>
--- a/Manifests/ManagedPreferencesApplications/com.anthropic.claudefordesktop.plist
+++ b/Manifests/ManagedPreferencesApplications/com.anthropic.claudefordesktop.plist
@@ -134,7 +134,7 @@
 			<key>pfm_description</key>
 			<string>Forces users to sign in with the specified Anthropic organization. When set, Claude Desktop requires the user to authenticate with the matching organization UUID.</string>
 			<key>pfm_documentation_url</key>
-			<string>https://support.claude.com/en/articles/12622667-enterprise-configuration</string>
+			<string>https://code.claude.com/docs/en/settings</string>
 			<key>pfm_format</key>
 			<string>^[0-9A-Za-z]{8}-[0-9A-Za-z]{4}-[0-9A-Za-z]{4}-[0-9A-Za-z]{4}-[0-9A-Za-z]{12}$</string>
 			<key>pfm_name</key>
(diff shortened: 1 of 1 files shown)


Manifests/ManagedPreferencesApplications/com.anthropic.claudefordesktop.plist
```diff
@@ -130,6 +130,22 @@
 			<key>pfm_type</key>
 			<string>string</string>
 		</dict>
+		<dict>
+			<key>pfm_description</key>
+			<string>Forces users to sign in with the specified Anthropic organization. When set, Claude Desktop requires the user to authenticate with the matching organization UUID.</string>
+			<key>pfm_documentation_url</key>
+			<string>https://support.claude.com/en/articles/12622667-enterprise-configuration</string>
```
relgit: I don't see `forceLoginOrgUUID` listed on this page. Could you update this to the correct one?

relgit (changes requested):
Thanks for the PR @rmalar.

Please see my sole comment in the code.

Manifests/ManagedPreferencesApplications/com.anthropic.claudefordesktop.plist
rmalary: So I actually noticed another company was using it and gave it a try, and with this custom profile in jamf 
Screenshot 2026-05-19 at 9 33 44 AM
the result is this 
Screenshot 2026-05-18 at 2 18 13 PM
The key is listed here https://code.claude.com/docs/en/settings. I noticed a lot of Anthropics documentation isn't up to date

relgit: I see it there. Use this link instead and then we can merge. Thanks.

kevinmcox:
I believe the new link (https://code.claude.com/docs/en/settings) applies to Claude code CLI, not Claude Desktop.

They are two different preference domains.

relgit:
Good call @kevinmcox. [Looks like](https://code.claude.com/docs/en/settings#settings-precedence) managed settings of Claude Code do pass through from Claude Desktop sometimes, but the emphasis is on **sometimes** 🙂

I suggest we move all Claude Code specific settings to a dedicated `com.anthropic.claudecode` manifest. If we mix and rely on the pass-through people might get frustrated when it's not working.

What do you think?

kevinmcox:
> Good call @kevinmcox. [Looks like](https://code.claude.com/docs/en/settings#settings-precedence) managed settings of Claude Code do pass through from Claude Desktop sometimes, but the emphasis is on **sometimes** 🙂
> 
> I suggest we move all Claude Code specific settings to a dedicated `com.anthropic.claudecode` manifest. If we mix and rely on the pass-through people might get frustrated when it's not working.
> 
> What do you think?

Absolutely we should have a separate manifest for code.

I leveraged our contract at work to get an official reply:

> Thanks for flagging. You're correct on both points: Claude Code and Claude Desktop are managed via separate preference domains (com.anthropic.claudecode vs com.anthropic.claudefordesktop), and forceLoginOrgUUID is a supported key in the Desktop domain. It was added to Desktop more recently than to Code, and the Desktop enterprise-configuration article hasn't caught up yet.
>
> To set expectations: the two products do not share an identical key set. Some Claude Code settings have Desktop equivalents and some don't, so I wouldn't assume every key from the Code settings page will be honored under the Desktop domain. forceLoginOrgUUID is one that is.
>
> I've passed your request to our docs team to get the article updated. Thanks again for the flag!

relgit:
Alright I'm going to change the name of this manifest and remove the doc URL before I merge because there's a number of people waiting on this now.

Then we can add the Code manifest separately.

PR #907, opened 2026-05-21
Add DockPilot managed preferences manifest

timr1985:
## Summary
- Add a DockPilot managed preferences manifest for the `mm.DockPilot` configuration domain.
- Include Dock policy, app management, smart replacements, right-side Dock items, and Dock appearance keys.
- Add the DockPilot icon and regenerate the manifest and icon indexes.

## Notes
- This manifest covers the DockPilot configuration payload only.
- Enterprise licence/enrolment keys are intentionally not included.

## Validation
- `plutil -lint Manifests/ManagedPreferencesApplications/mm.DockPilot.plist Manifests/index Icons/index`
- `git diff --check`


kevinmcox (changes requested):
Thanks for the contribution @timr1985!

You do not need to update the indexes, that happens automatically. Can you remove them from the PR please?

Also can you resize the icon to match others in the repo?

timr1985:
Thanks Kevin, updated now: removed the generated index changes from the PR and resized the DockPilot icon to match the existing 64x64 managed preference icons.

kevinmcox (changes requested):
Hey @timr1985, thanks for the updates — the icon resize and index removal both look good. The manifest itself is really nicely structured, especially the segmented controls for grouping the config areas. Nice work on a first submission.

  A few more things before this is ready to merge:

  **Drop the embedded `pfm_icon` data block.** The icon file you added in `Icons/ManagedPreferencesApplications/` is all we need — that's the standard pattern across the repo.

  **`PreserveUserItems` needs a bit of clarification.** The description says *"Prefer Allow User Dock Items for new profiles"* which makes it sound like a legacy key — if that's the case, could you add `pfm_app_deprecated` (the DockPilot version where it stopped being preferred) and `pfm_app_max` (last version where it actually works)? If it's still fully supported and just a stylistic preference, rewording the description to make that clearer would help admins understand which to pick.

  **Couple of nice-to-haves** (not blockers):
  - A `pfm_app_min` somewhere — either on the manifest as a whole or on individual keys — would help admins know which DockPilot version they need.
  - A few booleans (`Magnification`, `AutoHide`) and the size reals don't have `pfm_default` values. If DockPilot has documented defaults for them, worth adding; if "no value = let macOS handle it," that's fine to leave as-is.

  Thanks again!

timr1985:
Hi Kevin,

Thanks again for the clear review notes.

I’ve pushed the follow-up changes now:
- removed the embedded pfm_icon block
- kept the standalone icon file only
- clarified PreserveUserItems as a supported compatibility alias rather than a deprecated key
- added a manifest-level pfm_app_min
- added Dock appearance defaults where DockPilot has documented defaults
- confirmed the generated indexes are not included

Really appreciate the guidance on matching the repo patterns.

Thanks,
Tim

PR #909, opened 2026-05-22
Add com.trelica.macapp manifest

rderewianko:

**App Name**: Trelica Browser Helper

**App URL**: https://www.trelica.com/

**App Profile Documentation URL**: https://help.trelica.com/hc/en-us/articles/9946257398429-Deploying-on-macOS

**Payload / Domain**: com.trelica.macapp


kevinmcox (changes requested):
Hey! Thanks for the PR @rderewianko.

Couple of things:

**`Domain` shouldn't have `pfm_default`** — the docs make it clear there are two valid options (`app.trelica.com` for US, `eu.trelica.com` for EU), so `app.trelica.com` isn't really a default, it's just an example. Drop `pfm_default` and keep `pfm_value_placeholder` only. (Or even better, use a `pfm_range_list` with both valid domains.)

**Add `pfm_app_min` at the top of the manifest set to `1.3`** — the docs say v1.3 (May 2025) was the first version that supports MDM deployment with these `mcx_preference_settings`, so it's worth flagging that minimum version on the manifest.

**Throw a `pfm_value_placeholder` on `OrgId` too** — admins have no clue what format to expect when they hit this field. Something like the 32-character example from the install docs would help a lot.

Everything else looks great — the structure is clean, all the standard payload keys are there, and the icon's good. Cheers!

rderewianko:
Thanks for the feedback Kevin! I've adjusted this

relgit (approved):

PR #910, opened 2026-05-25
Update com.northpolesec.santa Manifest Keys

turquoise-turtle:
Added new keys, removed a few old keys that had been renamed or updated.

`ModeNotificationStandalone` was also added that was visible in https://github.com/northpolesec/santa/blob/main/Source/common/SNTConfigurator.h but not https://northpole.dev/configuration/keys/

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.northpolesec.santa.plist
+++ b/Manifests/ManagedPreferencesApplications/com.northpolesec.santa.plist
@@ -5,13 +5,13 @@
 	<key>pfm_description</key>
 	<string>Santa settings</string>
 	<key>pfm_documentation_url</key>
-	<string>https://northpole.dev/deployment/configuration.html</string>
+	<string>https://northpole.dev/configuration/keys/</string>
 	<key>pfm_domain</key>
 	<string>com.northpolesec.santa</string>
 	<key>pfm_format_version</key>
 	<integer>1</integer>
 	<key>pfm_last_modified</key>
-	<date>2025-01-26T23:49:44Z</date>
+	<date>2026-05-25T00:40:03Z</date>
 	<key>pfm_platforms</key>
 	<array>
 		<string>macOS</string>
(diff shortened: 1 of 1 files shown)


turquoise-turtle:
I kept running the pre-commit hook at various times but I might have forgotten before actually committing the changes, as I was testing on a different laptop.

I'm having trouble finding the right syntax for the dictionary type (for the MetricExtraLabels and SyncExtraHeaders keys I've added). 
https://northpole.dev/configuration/keys/#SyncExtraHeaders is a dictionary type that allows for custom key value pairs. I was only able to get it to show properly in iMazing Profile Editor if I used 
```
<key>pfm_name</key>
<string>SyncExtraHeaders</string>
<key>pfm_subkeys</key>
<array>
    <dict>
	    <key>pfm_type</key>
	    <string>string</string>
    </dict>
    <dict>
	    <key>pfm_type</key>
	    <string>string</string>
    </dict>
</array>
<key>pfm_title</key>
<string>Sync Extra Headers</string>
<key>pfm_type</key>
<string>dictionary</string>
```

Copilot had suggested this layout, but it would only show as asterisks as the title and blank string entry in iMazing Profile Editor, rather than the PLIST input. Which was why I removed the pfm_name keys.
```
<dict>
    <key>pfm_name</key>
    <string>*</string>
    <key>pfm_type</key>
    <string>string</string>
</dict>
```

relgit:
Hello @turquoise-turtle.

See https://github.com/ProfileManifests/ProfileManifests/wiki/User-Provided-Key-Names.

turquoise-turtle:
I think I've fixed up all the validation errors now. Please let me know if there's anything else I need to add

relgit (approved):

relgit:
Thanks @turquoise-turtle. You can tell lots of work went into this one.

turquoise-turtle:
Thanks!

PR #913, opened 2026-06-03
Elevate24 2.5 Update

liarrmo:
Elevate24 2.5 Update. 

Updates to com.jigsaw24.Elevate24.plist & com.jigsaw24.ElevateSecurityLogs.plist

Moved a few bits around for structure and added new keys. 

I've deleted two keys entirely, this is expected and want these fully removed. These are - 				
<string>EnableArchive</string>
<string>MaxArchiveAge</string>

Thanks

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.jigsaw24.Elevate24.plist
+++ b/Manifests/ManagedPreferencesApplications/com.jigsaw24.Elevate24.plist
@@ -113,7 +113,7 @@
 	RK5CYII=
 	</data>
 	<key>pfm_last_modified</key>
-	<date>2026-02-26T12:00:00Z</date>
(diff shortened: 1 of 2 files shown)


relgit:
Hi @liarrmo. Thanks for the PR.

We normally keep deprecated and removed properties, and mark them as such using `pfm_app_deprecated` and in `pfm_description` if needed. We do this so users could load existing profiles with no issues.

Do you mind re-adding the removed properties?

liarrmo:
Thanks @relgit! I'd like to just remove these keys, it causes issues with previous versions and we've advised our customers not to use it. So I'm happy to just remove it if possible. It's a legacy key that isn't being used, so need to remove references to it. 

Hope this makes sense

relgit (approved):

PR #914, opened 2026-06-04
Add manifest for com.microsoft.clawpilot (Microsoft Scout)

Haystack155:
Summary
Adds a preference manifest for the `com.microsoft.clawpilot` domain
(Microsoft Scout — macOS tenant policy controls).

Source
Keys and types derived from Microsoft's published configuration profile:
https://github.com/microsoft/scout-resources/blob/main/admins/microsoft-scout.mobileconfig

Keys included
PolicyVersion, ForcePrompt, DisableHeartbeat, DisableWorkflows,
RestrictToWorkspace, AllowScoutFrontierAccess, DisabledServers,
DisabledModels, DisabledPermissions, DisabledProviders,
BrowserEgressBlockedOrigins

Testing
- Loaded locally in iMazing Profile Editor via the local manifest
  override folder; all keys render with correct types and descriptions.


kevinmcox (changes requested):
Hey @Haystack155, thanks for the contribution! The structure is mostly there — a few things to fix before this can land:

**Drop `Manifests/index` from the PR.** That file regenerates automatically when changes hit master, so contributors don't need to touch it.

**`pfm_targets` values aren't right.** You've got `user-managed` and `system-managed`, but the manifest format only knows about `system` and `user` — anything else won't be recognized by ProfileCreator or iMazing. Worth a quick fix:

```xml
<key>pfm_targets</key>
<array>
    <string>system</string>
    <string>user</string>
</array>
```

**A few required top-level keys are missing.** Most app manifests in the repo declare these — could you add them?
- `pfm_unique` (almost certainly `true` here — only one Scout policy per profile)
- `pfm_app_url` (pointing to Scout's product page or the source repo)
- `pfm_documentation_url` (the `github.com/microsoft/scout-resources` link you cited in the PR description would work great)
- `pfm_platforms` (just `<string>macOS</string>` in an array)

**The standard payload subkeys are missing a few bits of polish.** Compare to other recent manifests in the repo for reference, but the gist:
- `PayloadIdentifier`, `PayloadUUID`, and `PayloadVersion` should each have `pfm_require` set to `always`
- `PayloadUUID` usually has a `pfm_format` regex to validate the UUID shape: `^[0-9A-Za-z]{8}-[0-9A-Za-z]{4}-[0-9A-Za-z]{4}-[0-9A-Za-z]{4}-[0-9A-Za-z]{12}$`
- `PayloadVersion` typically has `pfm_range_list` with `[1]`

**One thing I noticed but want to call out as intentional:** your boolean policy keys (`ForcePrompt`, `DisableHeartbeat`, etc.) don't have `pfm_default` values, and that's actually correct — the source mobileconfig comments make it clear that absent keys mean "not enforced," so no default is the right call. Just flagging so you don't second-guess it later.

**Style nits (optional, but easy):**
- Looks like you used 4-space indentation while the rest of the repo uses tabs. Running the file through `plutil -convert xml1 Manifests/ManagedPreferencesApplications/com.microsoft.clawpilot.plist` will normalize it.
- An icon at `Icons/ManagedPreferencesApplications/com.microsoft.clawpilot.png` (64×64) would be a nice finishing touch if you can grab Scout's product icon.

Thanks again — looking forward to seeing this land once those are sorted!

Manifests/ManagedPreferencesApplications/com.microsoft.clawpilot.plist
```diff
@@ -0,0 +1,307 @@
...
+			<string>PolicyVersion</string>
+			<key>pfm_title</key>
+			<string>Policy Version</string>
+			<key>pfm_description</key>
+			<string>Schema version. Set to 1.</string>
+			<key>pfm_type</key>
+			<string>integer</string>
+			<key>pfm_default</key>
+			<integer>1</integer>
+		</dict>
```
wjk, line 130: Should this be marked as required? (I never have used Scout, so I don’t know if this is truly required, but the description makes it sound as if it is.)

```xml
<key>pfm_required</key>
<true/>
```

Haystack155: You are correct thank you for pointing that out!

Haystack155 (commented):
Formatting and required key changes made.

PR #915, opened 2026-06-05
Update forceLoginOrgUUID to support array in com.anthropic.claudefordesktop

rmalary:
## What

Updates the `forceLoginOrgUUID` key in the Claude Desktop manifest (`com.anthropic.claudefordesktop`) to reflect that it accepts either a single UUID string **or** an array of UUIDs, as documented in the [Anthropic enterprise configuration docs](https://support.claude.com/en/articles/12622667-enterprise-configuration-for-claude-desktop).

## Why

The key was previously typed as `string`. Anthropic's documentation now clarifies it supports two modes:

- **Single string** — accepts one org UUID and pre-selects that organization at login
- **Array of strings** — accepts multiple org UUIDs; any listed org is accepted, but no org is pre-selected

## Changes

- `pfm_type`: `string` → `array` with UUID-format-validated string subkeys
- `pfm_title`: updated to `Force Login Organization UUID(s)` to reflect both forms
- `pfm_description`: rewritten to match Anthropic's official description of both behaviors
- `pfm_last_modified`: updated to 2026-06-06
- `pfm_version`: bumped to 6
- `Manifests/index`: updated accordingly

Made with [Cursor](https://cursor.com)


Manifests/ManagedPreferencesApplications/com.anthropic.claudefordesktop.plist
```diff
@@ -447,6 +456,6 @@
 	<key>pfm_unique</key>
 	<true/>
 	<key>pfm_version</key>
-	<integer>4</integer>
+	<integer>6</integer>
```
relgit: Please only increase the version to 5.

Manifests/index
```diff
@@ -370,11 +370,11 @@
 		<key>com.anthropic.claudefordesktop</key>
 		<dict>
 			<key>modified</key>
-			<date>2026-05-18T00:00:00Z</date>
```
relgit: There's no need for index updates – the index is generated automatically.

relgit (changes requested):
Thanks for the PR @rmalary !

One of these days we'll have to add multi-type support to PFM somehow because clearly we're not going to educate all devs that there is no key name shortage.

Anyway, please see and address my two comments. Cheers 🙂

relgit (approved):

PR #919, opened 2026-06-15
Update keys that were missing from segments

### LINKED ISSUE #590: Add allowRapidSecurityResponse keys to Restrictions payload
**Name of App / Payload**:
Restrictions: `com.apple.applicationaccess`

**Preference key(s) to be added / updated**:
`allowRapidSecurityResponseInstallation`
boolean
Default: true

`allowRapidSecurityResponseRemoval`
boolean
Default: true

**Applicable Preference Documentation URL(s)**:
https://developer.apple.com/documentation/devicemanagement/restrictions

turquoise-turtle:
Fixes #590 

Not finished yet, still working through some more manifests.


turquoise-turtle:
I think these changes are ready to be reviewed. The main goal is to just add the keys into the segments arrays so that it doesn't show in an "Other" category but in the appropriate one.

relgit:
Thanks for the PR @turquoise-turtle.

There's good work here but also some work that I think is not necessary, so let me explain.

In ProfileCreator, properties that are not assigned to a segment appear always. Sometimes above a segmented control, and sometimes below. This is intentional because segmented controls can nest inside others, and properties should be allowed to appear regardless of the selected segment.

For example, in the Enterprise Connect, I think it was the original intention for `adRealm` to be very prominent and to appear above the segmented control and regardless of what segment is selected.

iMazing Profile Editor on the other hand groups all properties into segments, and where no segment is assigned, drops properties into the 'Other' bucket. For `adRealm` putting it in 'Other' actually gives it the least prominence.

To address this incompatibility between the editors at iMazing, we manually manage the positions of properties in our editor, and in the case of Enterprise Connect this is clearly an omission which we will fix.

However, in other places you made changes, there is already manual management of segments, and properties don't get added to 'Other'. These cases should stay as they are today to preserve the ProfileCreator intended layout.

The following are the files in which your changes should be kept:
com.agilebits.onepassword7.plist
com.jamf.connect.login-Okta.plist
com.jamf.connect.login-OpenID.plist
com.jamf.connect.sync.plist
com.apple.applicationaccess-iOS.plist
com.apple.applicationaccess-macOS.plist

The others should be reverted.

Now, if you still think that the ProfileCreator layout, though intentional, does not make sense anymore, I will happily accept such an argument 🙂

turquoise-turtle:
Ah I see. I guess that does make sense in terms of the origins of this project, but I'll keep my preference of using iMazing Profile Editor over ProfileCreator 😄. I was actually basing some of my changes off how it looked to me in iPE of putting keys into a 'General' segment.

In the future I'll rework the script that helped identify these gaps to accept keys without segments, and focus more on the chore of spelling/capitalisation checks. I'm guessing I would look at adding it to some pre-commit repo instead of here?

I've reverted the changes now for those other files, except the `MessagePluginDisabled` typo still in `menu.nomad.NoMADPro.plist`

relgit:
Great – thanks!

The pre-commit script we're using is [here](https://github.com/homebysix/pre-commit-macadmin/blob/main/pre_commit_macadmin_hooks/check_preference_manifests.py) if you want to add to it.

PR #934, opened 2026-07-13
Update com.dare.zappl.preferences manifest for Zappl v3

dellis-dare:
Updates the Zappl (`com.dare.zappl.preferences`) manifest to match the v3 release of Zappl.

### GeneralOptions
- Rename `hourlyCheck` → `backgroundUpdates`
- Rename `backgroundUpdateWindow` → `downloadStaggerWindow` (conditional target repointed; existing `pfm_app_min` retained)
- Add `enableLiquidGlass` (Liquid Glass visual style on macOS 26 Tahoe+)

### ScheduledUpdates & ForcedUpdates
- Add info-button keys: `showInfoButton`, `initialPromptInfoPopover`, `gracePeriodPromptInfoPopover` (popover-text keys gated on `showInfoButton`)
- Convert `updateProgressPrompt` from a 3-value string enum to a boolean toggle; progress position/overlay-icon keys now gated on it
- Remove keys dropped in v3 (custom prompt-body messages, custom prompt sizes, Self Service skip button)

### Housekeeping
- Bump `pfm_version` to 3 and refresh `pfm_last_modified`

Validated with `plutil -lint` and the `check-preference-manifests` pre-commit hook (both pass).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Manifests/ManagedPreferencesApplications/com.dare.zappl.preferences.plist
+++ b/Manifests/ManagedPreferencesApplications/com.dare.zappl.preferences.plist
@@ -13,7 +13,7 @@
 	<key>pfm_format_version</key>
 	<integer>1</integer>
 	<key>pfm_last_modified</key>
-	<date>2026-07-14T10:36:58Z</date>
+	<date>2026-07-15T09:13:58Z</date>
 	<key>pfm_platforms</key>
 	<array>
 		<string>macOS</string>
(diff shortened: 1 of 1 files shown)


relgit (changes requested):
Thanks for the PR @dellis-dare.

I see there is quite a number of properties that were removed or renamed. We generally ask that these will be kept, so that users that load existing profiles with the old keys will have the best user experience.

Please add back any properties that were removed and mark them with `pfm_app_deprecated`.

As for properties that were renamed, please duplicate these and restore the original name on one of the instances (along with a deprecation tag as well).

Cheers 😀

dellis-dare:
> Please add back any properties that were removed and mark them with `pfm_app_deprecated`.
> 
> As for properties that were renamed, please duplicate these and restore the original name on one of the instances (along with a deprecation tag as well).

Thanks @relgit, much appreciated. Hopefully that's all sorted now.

One thing I did want to flag was our `updateProgressPrompt` property, as in v3 it changes type (list > boolean), but the property name stays identical. Would there be a best practice method for handling this that you could suggest? I haven't made any changes outside of altering the description to reflect the type change.
If additional context helps at all, Zappl v3 is backwards compatible with older properties such as this one.

relgit:
Thanks for making the changes, and so promptly 😀

The PFM format does not support more than one data type per property so there is no way to retain the old data type along the new one. The only best practice I can offer is to pick a new property name for such changes in the future.

You're in good company though, as I recently seen the [same thing](https://github.com/ProfileManifests/ProfileManifests/pull/915#pullrequestreview-4526148696) with Claude Desktop.

There's one more thing I want to address, and that's the changes you made to the descriptions of removed keys, where you prepended the following:

> (Deprecated in Zappl v3.0.0 — no longer used in Zappl v3 and later; retained so existing profiles continue to load.) 

I think that's a little too verbose. Personally I'd remove the whole thing since ProfileCreator already exposes `pfm_app_deprecated` and iMazing Profile Editor will too eventually.

But, if you still wish to keep something, I'd stick to a shortened "Deprecated in Zappl 3.0.0". You can also move these properties down to the bottom of the manifest to reduce their prominence.

Other than this, we're ready to merge.

dellis-dare:
Appreciate the guidance, definitely makes sense.

It's good to know that iMazing will also visually show deprecated keys in the UI, that was actually the main reason for the descriptions - The latest commits utilise `pfm_note` for the shortened deprecation callouts, and deprecated properties are now at the end of their respective property group.

Thanks again!

relgit:
Excellent.

PR #938, opened 2026-08-01
Add Missive

YesThatAllen:
Missive is a collaboration tool we use, see 
https://missiveapp.com/

Their development team implemented a means to prevent updating, by setting `AutoUpdate` to `false`

This PR formalizes the key, I'll update the PR with a proper documentation URL when possible.


kevinmcox:
> I'll update the PR with a proper documentation URL when possible.

Thanks @YesThatAllen! Do you want to hold the merge until you get the key, or add it in a separate PR later?

YesThatAllen:
In a bit of a catch-22 scenario, Missive 11.29.1 hasn't been released yet. Let's let this PR wait a bit. I just put this PR in draft mode and will update when this is actually released.

YesThatAllen:
Missive 11.33.0 is available for download, and the docs are updated, we should be good to go here.

Manifests/ManagedPreferencesApplications/com.missiveapp.osx.plist
```diff
@@ -0,0 +1,152 @@
...
+			<key>pfm_title</key>
+			<string>Payload Organization</string>
+			<key>pfm_type</key>
+			<string>string</string>
+		</dict>
+		<dict>
+			<key>pfm_default</key>
+			<true/>
+			<key>pfm_description</key>
+			<string>Prevents Missive from automatically checking for update to the Electron core. (Users will still see periodic request to "update" via reload.)</string>
```
kevinmcox: Two small changes:

> Prevents Missive from automatically checking for update**s** to the Electron core. (Users will still see periodic request**s** to "update" via reload.)

kevinmcox (changes requested):

Manifests/ManagedPreferencesApplications/com.missiveapp.osx.plist
YesThatAllen: Good catch

PR #940, opened 2026-08-02
Add more app icons

YesThatAllen:
While making some new manifests, I noticed ProfileCreator was missing some icons. 

I grabbed some, mocked up one for confboard, and it seems like vpntracker's name didn't line up


YesThatAllen:
Similarly, what do we think about having an icon like this for all the command line tools?

binary-tool

YesThatAllen:
Yeah, the more I thought about it, the better it seemed to have a proper icon instead of a generic missing image.

By having individual copies of the generic `exec` icon, it would be easy for someone to replace the image if any given tool adopts a logo.

kevinmcox (approved):
Yep I like the idea of having an icon, even if generic, versus nothing at all.

PR #942, opened 2026-08-07
Update links and two headings in the Zoom Desktop profile

### LINKED ISSUE #939: Review double "Enable Updates" in zoom's profile
The current zoom config has two keys named `Enable Updates` 

This one looks like it could have a better name

https://github.com/ProfileManifests/ProfileManifests/blob/5f77defe2b1c7ab491b989d18f650667a85abc8f/Manifests/ManagedPreferencesApplications/us.zoom.config.plist#L977-L989

Before I go making a PR to change the name for `ZAutoUpdate` to call it something like `Allow Manual Update Checks`

Is there a convention I'd be breaking with such?

YesThatAllen:
Closes https://github.com/ProfileManifests/ProfileManifests/issues/939 I may have more updates as I start digging in

### DIFF
--- a/Manifests/ManagedPreferencesApplications/us.zoom.config.plist
+++ b/Manifests/ManagedPreferencesApplications/us.zoom.config.plist
@@ -7,13 +7,13 @@
 	<key>pfm_description</key>
 	<string>Zoom settings</string>
 	<key>pfm_documentation_url</key>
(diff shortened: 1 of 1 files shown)


Manifests/ManagedPreferencesApplications/us.zoom.config.plist
YesThatAllen: This would make sense if it were true by default, however I can't really test this, as it seems the key is deprecated

```suggestion
			<true/>
```

kevinmcox (approved):

PR #961, opened 2026-09-13
Add manifest and icon for ChatGPT / Codex (com.openai.codex)

tarun101:
Adds a preference manifest and icon for the `com.openai.codex` domain, which is shared by the ChatGPT desktop app for macOS (bundle identifier `com.openai.codex`, currently 26.908.x) and the Codex CLI.

Codex reads two managed preference keys from this domain, both base64-encoded TOML strings:

- `config_toml_base64` — managed defaults. Applied above `/etc/codex/managed_config.toml` and the user's `config.toml`. Supported since Codex CLI 0.45.0.
- `requirements_toml_base64` — enforced requirements (for example `allowed_approval_policies`). Applied below the system `requirements.toml` and cloud-managed requirements. Supported since Codex CLI 0.78.0.

Source of truth for the key names and domain is the Codex macOS loader: https://github.com/openai/codex/blob/main/codex-rs/config/src/loader/macos.rs

Documentation: https://developers.openai.com/codex/enterprise/managed-configuration

Codex reads the values through `CFPreferencesCopyAppValue`, so `pfm_targets` is set to both system and user. `pfm_app_min` is intentionally omitted because the CLI (0.x) and the desktop app (26.x) use different version schemes.

Icon extracted from the app bundle at 64x64 RGBA. Manifest passes `check-preference-manifests` at the pinned pre-commit revision.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wjk:
Question. I know that, at least in iMazing Profile Manager, it is possible to have the manifest configured such that instead of having to paste large amounts of Base64 into a small text box, there can be a Choose button to select a file through the GUI, and the Base64 conversion is done automatically. Is there any reason why we can’t do that here?

relgit:
@wjk you're thinking about binary properties. Those are represented in XML plist as a Base64 strings inside `<data>` tags.

Whereas here it seems that the properties are string properties. In the XML that would be Base64 strings inside `<string>` tags.

Slight difference, I know, but meaningful. Only the former case is currently supported in the editors.

